PR #393, opened 2014-12-27
Add LocaleTrait to reduce code duplication

### LINKED ISSUE #378: Add LocaleTrait
I found myself copying/pasting code related to locale (basically, `$locale` attr, `getLocale()` and `setLocale()`). I think it could a nice use case for adding a `trait`. WDYT?

giosh94mhz:
This fix #378 

A couple of notes:
- I've kept "return $this" from setLocale, but it seems odd in a traits;
- I've copied the DocBlock from LocaleAwareProvider, but I don't know if it is the best way for the documentation.

Hope this helps :)


toin0u:
:+1:

PR #396, opened 2015-01-06
Fix encoding issue for maxmind provider

toin0u:
This is related to https://github.com/geocoder-php/BazingaGeocoderBundle/issues/84 but a patch should be written for the `2.x` version as well.


toin0u:
@willdurand no problem :) Thanks !

PR #397, opened 2015-01-06
Fix encoding issue for maxmind provider branch 2.x

toin0u:
This will fix the encoding issue https://github.com/geocoder-php/BazingaGeocoderBundle/issues/84


toin0u:
:+1:

Issue #392, opened 2014-12-27
[RFC] Use numbered administrative level instead of named one


giosh94mhz:
I know this may be a snowball in hell, since you are already in alpha stage, but I think I should give it a try before 3.0.0 is out :)

The Address model class, is actually defined to use "named" administrative level, so It has the following methods:
```php
final class Address {
    // [...]
    public function getCounty()
    public function getCountyCode()
    public function getRegion()
    public function getRegionCode()
    // [...]
}
```
The names seems ok at first and may be comfortable for some locations, but is pretty weird in many situation. A full table can be found on [Wikipedia](http://en.wikipedia.org/wiki/Table_of_administrative_divisions_by_country), but here some example:
* Botswana: District, Sub-District;
* China: Province, Prefecture, County, Township;
* France: Region, Departments, Arrondissements, Cantons;
* Italy: Region, Province;
* Mexico: State/District, Boroughs/Municipality;
* Montenegro: Municipality;
* Switzerland: Canton, District;
* USA: State, County;
* UK: Region, County;
* Vietnam: Province, District.

As you can see, using Region and County as subdivision is really weird in many situation. In China for example, the County is the 3rd admin level, but to be consistent across all Providers, if I use the method getCounty on a Chinese result I will get a Prefecture.

Also there is no benefit in having named division. The only affected user is the developer (since is at API level), but since this naming convention apply only to a very small subset of countries, it is not perceived as a feature (IMO). When a developer user geocoding it may be more comfortable with an interface which have "first admin level sub-division", "second level subdivision" and so on, so it simplify the process of creating hierarchies.

This is how is implemented by GoogleMaps and Geonames, as you can see from the Provider implementations.

If the interface is changed in this way, the classes Region and County are no longer needed and can be replaced by a single class "AdminLevel" (or many Admin1, Admin2, Admin3...). Also, providers can be implemented so that an array of Admin level returned which may improve the results generated by the Providers: in the current implementation there is no way of returning the 3rd and 4th admin level of a result in China, but If an array is returned there is no limit.

PS: I can create a PR, but first I'd like to hear your feedback.

toin0u:
@giosh94mhz Very interesting - I like it as well!

willdurand:
So I guess we should ship this into `alpha3`

PR #398, opened 2015-01-12
Use numbered administrative level instead of named one

giosh94mhz:
PR for for ticket #392. To sum up: replace every occurrence of `region` and `county` (no "r"), with a generic `admin` level array.

TODOs:

- [x] delete Region and County from Model namespace;
- [x] add Admin to Model namespace;
- [x] update failing test for Providers;
- [x] fix all providers code;
- [x] update string formatter;
- [x] update documentation;
- [x] squash all commit :)
- [x] update all tests (USE_CACHED_RESPONSES=false)

I'll like to hear you feedback about:

1. [minor] In `AbstractProvider::getDefaults` the admins value defaults to `[]` to simplify a bit the code. Is it ok?
2. Is `Admin` a clear class name or it is better to call it `AdminLevel` (as in #392)?
3. I'm using a simple array for admin levels, where the index is the "depth" in the hierarchy. Is this enough for you?
4. string formatter is now using `%R` and `%P` for `Region` and `County`, but this no longer make sense. What should be used? `%A1`, `%A2`,`%A3`,`%A4`? `%A0`, `%A1`, `%A2`,`%A3` (as the array index, not the admin level)? `%X`, `%Y`,`%Z`,`%W` (as in math :)? Something else?


willdurand:
> [minor] In AbstractProvider::getDefaults the admins value defaults to [] to simplify a bit the code. Is it ok?

Yes.

> Is Admin a clear class name or it is better to call it AdminLevel (as in #392)?

AdminLevel

> I'm using a simple array for admin levels, where the index is the "depth" in the hierarchy. Is this enough for you?

I am fine with it. I guess. Can we restrict this depth or it really depends on the location?

> string formatter is now using %R and %P for Region and County, but this no longer make sense. What should be used? %A1, %A2,%A3,%A4? %A0, %A1, %A2,%A3 (as the array index, not the admin level)? %X, %Y,%Z,%W (as in math :)? Something else?

%A1, %A2 sounds good to me.

poke @toin0u @Baachi

giosh94mhz:
As far as I know (i.e. the Geoname database), the depth is at most 4, and in some rare case is 5. Geonames doesn't support admin level 5 very well, so I suspect is very rare. For now, I'm just migrating the providers code so the maximum level for is 2. I may throw an exception if the admin depth is above 5 and we are safe, though.

I'll rename the class to `AdminLevel` ASAP.

(PS: I've fixed the mail address of my commit in the message above)

tests/Geocoder/Tests/Model/AddressFactoryTest.php
```diff
@@ -34,9 +34,9 @@ public function testCreateFromArray()
         foreach ($addresses as $address) {
             $this->assertInstanceOf('Geocoder\Model\Address', $address);
             $this->assertNull($address->getCoordinates());
-            $this->assertInstanceOf('Geocoder\Model\County', $address->getCounty());
+            $this->assertInstanceOf('Geocoder\Model\County', $address->getAdmins()[1]);
```
pyrech: I think the assertion is wrong, it should assert than it's an instance of `Geocoder\Model\Admin` (and `AdminLevel` after you change it) ;)

pyrech:
Small proposition: What about adding a `$level` property to `Admin`?

Also, currently the `Address#admins` array is 0-based and it looks like admin levels usually take values from 1 to 10 (eg on [OpenStreetMap](http://wiki.openstreetmap.org/wiki/Tag:boundary%3Dadministrative#admin_level)). Should the implementation match this behaviour?

tests/Geocoder/Tests/Model/AddressFactoryTest.php
giosh94mhz: Sure, I've started backward alphabetical order when fixing tests :) I have failing tests from A-GeoIP.

giosh94mhz:
@pyrech I think that the idea of having `$level` property is good!

I;ve looked into OpenStreetMap docs you supplied, but the value they expose is not the "administrative level" strictly speaking. As you can see, table data are sparse; in particular:
* level 1 and 2 are to be stripped from admins, and set to `Address::$country`
* level `n` and `n-1` (ignoring null values) are `City` and `SubUrbs`, which must be set to `Address::$city` and `Address::$subLocality`
* 3..(n-2): the remaining are at most 6 administrative level, but still seems very sparse and can be reduced to 4-5 (need to dig in the docs more).

IMO, the OpenStreetMap approach is too sparse, and must be aligned to something like Geonames or GMap (e.g [maps.api/?country=Spain&postalCode=08001](http://maps.googleapis.com/maps/api/geocode/json?components=country:spain|postal_code:08001&v=3&sensor=false&language=en)).

pyrech:
@giosh94mhz Yes, you're completely right about their structure. I was more refering to the fact they start to the level 1 instead of 0 and the GMap example is indeed far better.

Do you think that `Address#admins` should use real level as key? IMO `$address->getAdmins()[X-1]` doesn't look nice to get the administrative area level X.

giosh94mhz:
@pyrech Ok, now I got it :)

I though that it was a bit ugly the first time I wrote `getAdmins()[0]` :) . Anyway, using `$address->getAdmins()[X]` instead of `X-1` seems ok to me, probably is even better.

An alternative may be to have a syntactic sugar like:
```php
public function getAdmins()
{
    return $this->admins;
}
public function getAdmin($level)
{
    return isset($this->admins[$level-1]) ? $this->admins[$level-1] : null /* or maybe null obj if $level <= 4 or 5 */;
}
```

pyrech:
@giosh94mhz Hehe, we came to the same conclusion (I edited my previous comment in the same time)! :)

giosh94mhz:
Ok, I've renamed `Admin` to `AdminLevel` everywhere... it took me a while.

I've also used `1` based arrays, and they feel more natural when accessing specific level, but they are sometimes boring to write in provider (e.g. `[1 => ['name' => 'admin level']]`).

I'm trying to forcing a limit to admin level [1,5] wherever possible. I'm also checking the api docs of some provider, to see if there is support for multiple admin levels (2+).

Along the way I find some minor bugs as in commit 91b2e0d. @willdurand may I add it here (and not squashing in the end), or you prefer me to create another PR?

I'll wait for your feedback about  `getAdminLevel($level)`, in the meanwhile I keep up the work with the providers.

src/Geocoder/Model/AddressFactory.php
```diff
@@ -20,6 +22,18 @@ public function createFromArray(array $results)
...
+            foreach ($this->readArrayValue($result, 'adminLevels') as $level => $adminLevel) {
+                $level = $this->normalizeLevelValue($level);
+                $adminLevels[$level] = new AdminLevel(
+                    $level,
+                    $this->readStringValue($adminLevel, 'name'),
+                    $this->readStringValue($adminLevel, 'code')
+                );
+            }
+
+            ksort($adminLevels, SORT_NUMERIC);
```
Baachi: Is this needed? I see no reason for this.

src/Geocoder/Model/AdminLevel.php
```diff
@@ -11,11 +11,16 @@
 namespace Geocoder\Model;
 
 /**
- * @author William Durand <william.durand1@gmail.com>
```
Baachi: Don't remove the `@author` tag please.

Baachi:
Really great work @giosh94mhz. 
> [minor] In AbstractProvider::getDefaults the admins value defaults to [] to simplify a bit the code. Is it ok?

Yes

> I'm using a simple array for admin levels, where the index is the "depth" in the hierarchy. Is this enough for you?

Its enough, but i would enforce a string cast, so PHP does not made any magic with this array.

> string formatter is now using %R and %P for Region and County, but this no longer make sense. What should be used? %A1, %A2,%A3,%A4? %A0, %A1, %A2,%A3 (as the array index, not the admin level)? %X, %Y,%Z,%W (as in math :)? Something else?

I would tend to %A1, %A2, %A3, ...

toin0u:
Nice work @giosh94mhz !!

> I'm using a simple array for admin levels [..]

We can maybe use a `AdminLevelCollection` like https://github.com/geocoder-php/Geocoder/blob/master/src/Geocoder/Model/AddressCollection.php ?

> string formatter is now using %R and %P for Region and County, but this no longer make sense. What should be used? %A1, %A2,%A3,%A4? %A0, %A1, %A2,%A3 (as the array index, not the admin level)? %X, %Y,%Z,%W (as in math :)? Something else?

I agree with @Baachi

willdurand:
+1 for the collection

+1 for adding commits to this PR, we will see how to deal with all the commits later

willdurand:
thank you for making this real!

giosh94mhz:
Thanks @Baachi, @toin0u and @willdurand for your feedback and support!

> Its enough, but I would enforce a string cast, so PHP does not made any magic with this array.

PHP is full of magic, which one in particular? :) String are not so comfortable to use, since I'm actually dealing with numbers and have some checks for array bounds.

> We can maybe use a AdminLevelCollection like https://github.com/geocoder-php/Geocoder/blob/master/src/Geocoder/Model/AddressCollection.php ?

I'm also :+1: for this, now that I think of it.
The `AdminLevelCollection` should be:
* sorted by depth, so that the results can be iterated easily;
* key (i.e. level) are within [1,5], throw OutOfBound otherwise;
* if two different AdminLevel instance are added for the same level (during contruction at least) an InvalidArgument must be thrown.
This way, some checks are removed from the factory, which didn't fit well.

I'll keep index-based array for the internal result data.

src/Geocoder/Model/AdminLevel.php
giosh94mhz: It's not removed, the IDE changed to my tag author during the refactoring. I'll revert it anyway, no problem.

src/Geocoder/Model/AddressFactory.php
giosh94mhz: I think I've added a test which expose this, but here the PHP magic to solve:
```PHP
$admins = [];
$admins[2] = 'admin 2';
$admins[1] = 'admin 1';

echo implode(', ', $admins) . "\n";
foreach ($admins as $level => $admin) {
    echo $level . ': ' . $admin . "\n";
}
```

expected:
```
admin1, admin2
1: admin 1
2: admin 2
```

actual:
```
admin2, admin1
2: admin 2
1: admin 1
```

Since Providers deal with external data, I think is much safer to sort it out explicitely. Anyway, all of this is solved by introducing the `AdminLevelCollection` class.

Baachi: Ah cool, didn't see this test. Anyway i agree with you, this is more logical :+1:

Baachi:
> PHP is full of magic, which one in particular? :) String are not so comfortable to use, since I'm actually dealing with numbers and have some checks for array bounds.

I now, but i have some strange issues with it (i don't remember exactly which one). But if we decide to introduce an `AdminLevelCollection` it will be safer for the user land anyways.

Baachi:
I would tend to use the `SplFixedArray` which already throw an exception if more than x elements are given.

giosh94mhz:
@Baachi I didn't know about `SplFixedArray`.

Here's my thought
* PRO:
  * array is index-sorted as in C;
  * class definition is almost a 1-line alias of  `SplFixedArray`
* CONS:
  * since it's fixed, `count` is always 5, but countries may have less admin level (e.g. Italy got 2-3 level)
  * `null` value are interleaved with object value, so it can be unconfortable when looping
  * I need to catch and re-throw RuntimeException exceptins, if we want to have all exception in Geocoder namespace

So I think I'll create a new class and eventually try to use `SplFixedArray` as the inner instance.

What do you think?

Baachi:
The performance should be irrelevant. I would it only use if you have less code to write and its easier to maintain. If its more work, build your own :)

giosh94mhz:
From what I've read, `SplFixedArray` is meant to be (and seems to be) faster than regular array... anyway, for 5 element it doesn't matter.

This evening I'll work a bit more... I'll keep you up to date.

giosh94mhz:
Here I am! :) I've finished testing the providers, but I may need some help with testing which use api-keys... Hopefully travis will help.

About the `SplFixedArray`. I've tried integrating that class but is not very handy, because:

* it's read/write, not read-only so I need to wrap it anyway;
* count is always 5, no matter what;
* foreach always loop all 5 elements;
* index start from zero, so `$i == $level - 1`;
* interface is way different from `AddressCollection`;

Along the way, I have:

* improved Geonames and GoogleMap providers which now support all five admin levels. I'll try to do the same with the others.
* fixed the type hint for `Geocoder::geocode' method.

I think there is a more room for improvements. We may:

1. rename `*Collection::all` as  `*Collection::toArray` (which is "forward compatible" with DoctrineCommon and other collections, if you ever decide to switch), since it's more clear and the class itself is Traversable;
2. implement `ArrayAccess` interface in both collections (keeping the `get` method as an alias of `offsetGet` is good IMO);
3. if not implementing (2) add an `has` method, and throw when getting non existing keys.

willdurand:
> rename *Collection::all as *Collection::toArray (which is "forward compatible" with DoctrineCommon and other collections, if you ever decide to switch), since it's more clear and the class itself is Traversable;

:-1: I like Symfony "ParameterBag" API.

> implement ArrayAccess interface in both collections (keeping the get method as an alias of offsetGet is good IMO);

:-1: I'd like to keep collections read-only as much as possible

> if not implementing (2) add an has method, and throw when getting non existing keys.

What is the current impl of `get`?

giosh94mhz:
> I like Symfony "ParameterBag" API.
Ok, let's keep it like that.
> I'd like to keep collections read-only as much as possible
Me too. I've avoided read-write interface for this patch.
> What is the current impl of get?
`AddressCollection::get` return an address object or throw when index is greater than `count`.
But for `AdminLevelCollection` this may not be enough, since in (rare) situation there may be admin1 and admin3 level, without admin2 (i.e. count = 2, but admin3 is available). So I now return `OutOfBounds` when index is greater than 5 and `null` when the admin level is not available. The `has` method will simplify this implementation, and always throw when a level is not set.

giosh94mhz:
Ok, I've finally got all the api keys and tested all the providers with real data. I've fixed the commit history (all commit have passing tests), so that fixes I've fond along the way have a separate commit from the new admin feature.

Testing are now passing, but if I set `USE_CACHED_RESPONSES=false` some tests are failing. This is probably not related to my changes, but I'll try to fix them anyway.

@willdurand I've fixed the collection interfaces as in `ParametersBag`, please give it a quick review when you have the time.

toin0u:
Wow ! Amazing job ! I'll review it as well asap :)

toin0u:
:+1:

src/Geocoder/Formatter/StringFormatter.php
```diff
@@ -51,19 +49,26 @@ class StringFormatter
...
-            self::REGION        => $address->getRegion()->getName(),
-            self::REGION_CODE   => $address->getRegion()->getCode(),
             self::COUNTRY       => $address->getCountry()->getName(),
             self::COUNTRY_CODE  => $address->getCountry()->getCode(),
             self::TIMEZONE      => $address->getTimezone(),
-        ));
+        ];
+
+        $adminLevels = $address->getAdminLevels();
+        $nullAdminLevel = new AdminLevel(null, null, null);
```
willdurand: Should we return a null object or only null?

giosh94mhz: I don't understand what you mean, since the returned value is the formatted string.
Here I've used a null object just to avoid two cycle: the first to set all ADMIN_LEVEL_* `$tr` to null, and the second (foreach adminLevels) to set the actual value for the address. Since `AdminLevel` class is final there should be no problem.

willdurand: could you please avoid this null object?

giosh94mhz: Sure, no prob.

giosh94mhz:
Ok, I've made some fixes and improved Yandex/OpenStreetMap provider a bit.

I've changed the way cache hash is generated, so that it is independant of the api_key; by doing this, any developer can use the same cache or share an updated version.

I have some other minor API suggestion, but I think that I'll open a new PR :)

src/Geocoder/Provider/Yandex.php
```diff
@@ -122,27 +122,34 @@ function ($value, $key) use (&$details) {$details[$key] = $value;}
 
             if (! empty($details['lowerCorner'])) {
                 $coordinates = explode(' ', $details['lowerCorner']);
-                $bounds['south'] = $coordinates[1];
-                $bounds['west']  = $coordinates[0];
+                $bounds['south'] = +$coordinates[1];
```
toin0u: Why is a `+` necessary here ?

giosh94mhz: It is not strictly necessary.

I've seen that results for bounds were returned as numeric strings instead
of floating point. With the plus sign it is casted to float, as in
latitude, longitude, and other providers.
Il 06/feb/2015 00:09 "Antoine Corcy" <notifications@github.com> ha scritto:

willdurand: Sounds a bit hackish to me :p

toin0u: Maybe just cast with `(float)` ? :)

giosh94mhz: Maybe I'm biased from Javascript :)
I'll cast to float the good-old way.

giosh94mhz:
Rebased on master :)

src/Geocoder/Formatter/StringFormatter.php
```diff
@@ -51,19 +49,27 @@ class StringFormatter
...
-            self::COUNTY_CODE   => $address->getCounty()->getCode(),
-            self::REGION        => $address->getRegion()->getName(),
-            self::REGION_CODE   => $address->getRegion()->getCode(),
             self::COUNTRY       => $address->getCountry()->getName(),
             self::COUNTRY_CODE  => $address->getCountry()->getCode(),
             self::TIMEZONE      => $address->getTimezone(),
-        ));
+        ];
+
+        for ($level = 1; $level <= AdminLevelCollection::MAX_LEVEL_DEPTH; ++ $level) {
```
willdurand: would be nice to write `$level++`

src/Geocoder/Formatter/StringFormatter.php
```diff
@@ -51,19 +49,27 @@ class StringFormatter
...
+            $tr[self::ADMIN_LEVEL . $level] = null;
```
willdurand: what does `tr` stand for?

src/Geocoder/Provider/GeoIPs.php
```diff
@@ -165,20 +165,31 @@ private function executeQuery($query)
...
-            'longitude'   => '' === $location['longitude']    ? null : $location['longitude'],
-            'timezone'    => '' === $location['timezone']     ? null : $location['timezone'],
+        $location = array_map(function ($value) {
+            return '' === $value ? null : $value;
+        }, $response['location']);
+
+        $adminLevels = [];
+
+        if (null !== $location['region_name'] || null !== $location['region_code']) {
+            $adminLevels[] = ['name' => $location['region_name'], 'code' => $location['region_code'], 'level' => 1];
```
willdurand: could you write this line on multiple lines:

... [
    'name' => ...,
    'code'  => ...,
    ...
];

willdurand:
Last nitpicks. Time to get this merged!

src/Geocoder/Formatter/StringFormatter.php
giosh94mhz: Translate... called `tr` since the function using it is `strtr` :) I'll find a better name...

src/Geocoder/Formatter/StringFormatter.php
giosh94mhz: Of course. But don't tell  C++ developers I did it :P

giosh94mhz:
:+1: done!

willdurand:
\o/

toin0u:
@giosh94mhz Awesome!!! :fireworks:

willdurand:
https://github.com/geocoder-php/Geocoder/releases/tag/v3.0.0-alpha3

Baachi:
Really cool! :fireworks:

giosh94mhz:
Thank you everyone! You gave me great support and feedback!

PR #401, opened 2015-01-22
Fixed dependencies

GrahamCampbell:
NB, 0.x builds can break between minor releases so we must not use `~0.1` for a version constraint.


toin0u:
:+1:

willdurand:
thank you!

PR #402, opened 2015-01-23
Add optional "kind" parameter

tannermccoleman:
Add the parameter to specify the toponym type. More information on this parameter can be found [here](http://api.yandex.com/maps/doc/geocoder/desc/concepts/input_params.xml) 
(**Note**: This documentation claims it is only for *reverse geocoding*, but this is incorrect)

### DIFF
--- a/src/Geocoder/Provider/Yandex.php
+++ b/src/Geocoder/Provider/Yandex.php
@@ -52,7 +52,7 @@ class Yandex extends AbstractHttpProvider implements LocaleAwareProvider
     /**
      * {@inheritDoc}
      */
-    public function geocode($address)
+    public function geocode($address, $kind = null)
     {
         // This API doesn't handle IPs
         if (filter_var($address, FILTER_VALIDATE_IP)) {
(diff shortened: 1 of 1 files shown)


willdurand:
Unfortunately, you break the interface, which is not desired.

tannermccoleman:
Heh I realized that pretty much right after I submitted it. Would there be any way to implement this without having to unreasonably rework much? Adding the parameter is a night and day difference with Yandex's geocoding abilities, at least as far as Canadian addresses go.

toin0u:
The `toponym` is set in the constructor https://github.com/geocoder-php/Geocoder/blob/master/src/Geocoder/Provider/Yandex.php#L44

This is used here https://github.com/geocoder-php/Geocoder/blob/master/src/Geocoder/Provider/Yandex.php#L74-L76

I think you can move in the `Yandex::executeQuery` method :)

toin0u:
@tannermccoleman How can I test what that the Yandex documentation is incorrect? Do you have any example?

Issue #404, opened 2015-01-31
Documentation: Not clear how to provide the API key


gajus:
https://github.com/geocoder-php/Geocoder/blob/master/README.md `Geocoder\Provider\GoogleMaps` example fails to show that the last parameter is the API key. Had to take a look at the source code to find it, https://github.com/geocoder-php/Geocoder/blob/master/src/Geocoder/Provider/GoogleMaps.php#L56.

toin0u:
Thank you for the report @gajus :)

PR #406, opened 2015-02-01
Add the api key parameter to the google maps provider doc

toin0u:
Fix https://github.com/geocoder-php/Geocoder/issues/404


willdurand:
thanks!

PR #408, opened 2015-02-12
Allow locale change with chain

AntoineLemaire:
Need to change the locale dynamically in PHP for after use it with the CacheProvider of GeocoderBundle

### DIFF
--- a/src/Geocoder/Provider/Chain.php
+++ b/src/Geocoder/Provider/Chain.php
@@ -16,8 +16,10 @@ use Geocoder\Exception\InvalidCredentials;
 /**
  * @author Markus Bachmann <markus.bachmann@bachi.biz>
  */
-class Chain implements Provider
+class Chain implements LocaleAwareProvider
 {
+    use LocaleTrait;
+
     /**
      * @var Provider[]
      */
@@ -38,6 +40,10 @@ class Chain implements Provider
     {
         $exceptions = [];
         foreach ($this->providers as $provider) {
+            if ($provider instanceof LocaleAwareProvider && $this->getLocale() !== null){
+                $provider = clone $provider;
+                $provider->setLocale($this->getLocale());
+            }
             try {
                 return $provider->geocode($address);
             } catch (InvalidCredentials $e) {


src/Geocoder/Provider/Chain.php
```diff
@@ -38,6 +40,9 @@ public function geocode($address)
     {
         $exceptions = [];
         foreach ($this->providers as $provider) {
+            if($provider instanceof LocaleAwareProvider && $this->getLocale() !== null){
```
willdurand: Could you please use the coding standards of this project?

* 1 space before `(`
* `null !== ...` rather than the other way around

Does it make sense to check whether locale is null? It does not really matter IMO.

willdurand: What if registered providers are configured with a locale?
ping @Baachi

Baachi: @willdurand This should not be a problem, because the chainprovider delegate each call. 
Change the locale dynamicaly isn't supported yet. And in my opinion, such functionality made this code more complex.

If someone need this functionality, just create more than one provider like `provider_de, provider_en` and so on.
I think the `Geocoder` class should be decide which provider should be used not the provider itself. A provider should be invisible for the end user. And if someone _really_ need this functionality, here you go we are in OOP so extend the class.

AntoineLemaire: Problem is when using GeocoderBundle. 
For exemple, I got a default locale for Bing Map
```
providers:
    bing_maps:
       api_key:  ~
       locale:    fr_FR
```

I'd like some times set a different locale, but if not, it should use the default one, not erase the default by null

willdurand:
ah right, then fix CS and I'll merge this PR :)

AntoineLemaire:
Done ! Sorry for this mistake.

src/Geocoder/Provider/Chain.php
```diff
@@ -38,6 +40,9 @@ public function geocode($address)
     {
         $exceptions = [];
         foreach ($this->providers as $provider) {
+            if ($provider instanceof LocaleAwareProvider && $this->getLocale() !== null){
+                $provider->setLocale($this->getLocale());
```
Baachi: This can lead to an logical error, see this example:

```php
$bing = new BingMaps($http, 'secret', 'de_DE'); // note the locale here
$chain = new Chain([$bing]);

$geocoder = new Geocoder();
$geocoder->registerProvider($bing);
$geocoder->registerProvider($chain);

$geocoder->geocode('Paris'); // Will now use de_DE locale
$chain->setLocale('en_US') // Chain now use en_US
$geocoder->use('chain')->geocoder('Paris'); // I get en_US results

$geocoder->geocode('Paris'); // I expect the locale is de_DE, but it will be en_US because the chain provider will override them.
```

I would suggest to clone the provider or set the original locale back.

willdurand: :+1:

toin0u:
:+1:

willdurand:
@AntoineLemaire could you please squash your commits?

AntoineLemaire:
Done ! Thanks

Baachi:
Thank you very much for your contribution!

willdurand:
\o/

PR #423, opened 2015-04-14
improve Travis builds

xabbuh:
* use the faster container-based infrastructure
* remove unneeded `sh` wrapper
* don't specify default `--dev` argument
* run tests against the lowest possible dependency versions too


willdurand:
great, thanks!

Issue #431, opened 2015-05-07
Yandex provider: http transport is deprecated


nitso:
HTTPS should be used instead.
Official news: http://clubs.ya.ru/mapsapi/replies.xml?item_no=56858

Translation: 

> Safety is paramount, so all Yandex services are moving to the access scheme for HTTPS-only. Yandex not be an exception. Soon the JavaScript API, Static API and HTTP geocoder will be available only for HTTPS.
> Users of JS API, connecting on HTTP, will be redirect to HTTPS and API will continue to work as usual. Nevertheless, we recommend to reconnect API for HTTPS.
> If you are using HTTP geocoder, please make sure that the data is exchanged by HTTPS. Redirect Server may not occur, and your service can start working properly.

PR #432, opened 2015-05-07
Use https transport in Yandex provider

nitso:


Baachi:
You have to change the exception message in the unit tests.

willdurand:
Also, I would prefer to keep the `http` version. Some other providers have a `$useSsl` option.

toin0u:
I also think the `$useSsl` option is the best approach https://github.com/geocoder-php/Geocoder/blob/master/src%2FGeocoder%2FProvider%2FGoogleMaps.php#L80 :)

toin0u:
Oh Yandex will deprecate http https://github.com/geocoder-php/Geocoder/issues/431.. So this PR makes sense :) But like @Baachi said, tests should be fixed :)
- https://github.com/geocoder-php/Geocoder/blob/master/tests%2FGeocoder%2FTests%2FProvider%2FYandexTest.php#L41
- https://github.com/geocoder-php/Geocoder/blob/master/tests%2FGeocoder%2FTests%2FProvider%2FYandexTest.php#L51
- https://github.com/geocoder-php/Geocoder/blob/master/tests%2FGeocoder%2FTests%2FProvider%2FYandexTest.php#L61
- https://github.com/geocoder-php/Geocoder/blob/master/tests%2FGeocoder%2FTests%2FProvider%2FYandexTest.php#L71
- https://github.com/geocoder-php/Geocoder/blob/master/tests%2FGeocoder%2FTests%2FProvider%2FYandexTest.php#L81
- https://github.com/geocoder-php/Geocoder/blob/master/tests%2FGeocoder%2FTests%2FProvider%2FYandexTest.php#L249
- https://github.com/geocoder-php/Geocoder/blob/master/tests%2FGeocoder%2FTests%2FProvider%2FYandexTest.php#L259
- https://github.com/geocoder-php/Geocoder/blob/master/tests%2FGeocoder%2FTests%2FProvider%2FYandexTest.php#L269

nitso:
It is a good idea to implement ```$useSsl``` switch. Should it be true or false by default?

toin0u:
I'm not sure because Yandex will use https only soon..

Baachi:
I would rather throw an exception if the `openssl` extension isn't installed, because Yandex didn't support `http` longer.

willdurand:
Well if `http` is deprecated, then no need to introduce the `useSsl` option. However, it implies a sort of _soft_ BC break... I am not sure about how to deal with that..

nitso:
I've fixed tests that mentioned http. But there is still one failing. Can't really say why.

willdurand:
sounds good to me. could you please squash your commits?

nitso:
Done. 

I've added cached responses as well.

willdurand:
thank you very much!

PR #434, opened 2015-05-23
Update links to PSR-7 spec

miclf:
Since its acceptance, the PSR-7 document has been moved from proposed to accepted FIG standards. As a result, all the links that were targeting the proposed document became broken.

This PR fixes this by updating links in the README in order to make them reference the new location of PSR-7.


willdurand:
thank you!

PR #435, opened 2015-05-28
Made Google auth failures more verbose

amenk:


src/Geocoder/Provider/GoogleMaps.php
```diff
@@ -153,8 +153,8 @@ private function executeQuery($query)
             throw new NoResult(sprintf('Could not execute query "%s".', $query));
         }
 
-        if ('REQUEST_DENIED' === $json->status && 'The provided API key is invalid.' === $json->error_message) {
-            throw new InvalidCredentials(sprintf('API key is invalid %s', $query));
+        if ('REQUEST_DENIED' === $json->status) {
+            throw new InvalidCredentialsException(sprintf('API access denied. Request: %s - Message: %s', $query, $json->error_message));
```
Baachi: The class is `InvalidCredentials` and not `InvalidCredentialsException`

amenk: can you fix it? I was working with an old version of the code.

Baachi: I can't fix it, this is the new schema. You have to rename it back.

amenk: now it's all green

src/Geocoder/Provider/GoogleMaps.php
```diff
@@ -153,8 +153,9 @@ private function executeQuery($query)
             throw new NoResult(sprintf('Could not execute query "%s".', $query));
         }
 
-        if ('REQUEST_DENIED' === $json->status && 'The provided API key is invalid.' === $json->error_message) {
-            throw new InvalidCredentials(sprintf('API key is invalid %s', $query));
+        if ('REQUEST_DENIED' === $json->status) {
```
willdurand: I don't really get the point of this change. Could you explain why you removed the second condition?

willdurand: I mean, if the message has changed, then let's change the message in the code rather than removing it.

amenk: There are different messages with that code. I for example got a message that access from my IP is not allowed on the API key and the IP that google recognizes is dumped. So it is very helpful to include such messages in the exception.

willdurand: Ok, I see. However, the exception is thrown only when the request is denied because of credentials issue, which is not the case anymore with your patch.

amenk: okay get it - so we should throw another exception for those cases?

willdurand: yup

amenk:
I updated my PR

willdurand:
thanks!

Issue #417, opened 2015-03-18
GoogleMapsBusinessProvider not working 


webdesign7:
Hi, 

I can't get your geocode library  working  with GoogleMapsBusinessProvider .. every time it says:

Could not execute query http://maps.googleapis.com/maps/api/geocode/json?address=London%252C%2BUnited%2BKingdom&client=147436736XXX-pr4smhsrfvlgvb0ppp05unbv01akmc6k.apps.googleusercontent.com

my CODE:
        $geocoder = new \Geocoder\Geocoder();
        $adapter  = new \Geocoder\HttpAdapter\CurlHttpAdapter();
        $provider = new \Geocoder\Provider\GoogleMapsBusinessProvider($adapter,'14743673XXXX-nk3e8pubf834gk4ss555a6g8jg8sku12.apps.googleusercontent.com');
        $geocoder->registerProvider($provider);
        $response = $geocoder->geocode($location);

but the GoogleMapsProvider is working .. 

I am using the 2.8.2 version

I will appreciate your soon help as I am in the middle of the project working on this issue

Many thanks

webdesign7:
Hi there, 

Any help ? ideas, suggestions ? 

Thanks

toin0u:
@webdesign7 Have you look at this https://github.com/geocoder-php/Geocoder/issues/295 and this https://github.com/geocoder-php/Geocoder/issues/281 ?

dereuromark:
I had the same problem a few days ago, those screwups silently deactivated their API URL, you must use the primary one now - see https://github.com/dereuromark/cakephp-tools/commit/796864e49bd692bd0359e12bac5ff44298d75095

willdurand:
Anyone to fix the URL?

PR #436, opened 2015-06-23
Update url in GoogleMapsBusiness provider

elnoro:
Supposed to fix https://github.com/geocoder-php/Geocoder/issues/417


elnoro:
Interesting, seems to be a problem with curl https://bugs.launchpad.net/ubuntu/+source/gnutls26/+bug/1111882

PR #437, opened 2015-06-25
Add phpunit to composer.json

nitso:
Phpunit was added to dev section.


willdurand:
not sure it is useful, you can still install it globally.

PR #450, opened 2015-09-03
Fix emergency message in Geoip provider

misterion:
The `geoip_region_name_by_code()` and `geoip_time_zone_by_country_and_region()` require both arguments (country and region) should be provided. Code which previously served as a part of this verification has been removed 16.11.2012 (b4c48686) by Christaudo contributor. Now behaviour is wrong becouse of empty `region` in `geoip_record_by_name` very common situation and provide `You need to specify the country and region codes.` error message.

This pull request add this verification back.


willdurand:
thanks!

PR #456, opened 2015-10-02
Bugfix, Sometimes $data is null at this point.

Nyholm:
This fix removes the `E_WARNING: Invalid argument supplied for foreach()`


willdurand:
Heya! Do you think you could reveal this bug with a test case?

Nyholm:
Sorry for being lazy. Sure. I'll write something up this weekend.

Nyholm:
Updated. When we call the geocoder with `null` we use the following url: 
```
https://geocode-maps.yandex.ru/1.x/?format=json&geocode=&results=5
```
... which gives the following response:
```
{"error":{"status":"400","message":"missing geocode parameter"}}
```

This PR checks if there is an error.

toin0u:
:+1:

willdurand:
thanks!

Nyholm:
Thank you for merging.

PR #465, opened 2015-10-20
Add \Geocoder\Provider\Yandex::setToponym()

bubnovKelnik:


Baachi:
Please give a use case why do you need this.

bubnovKelnik:
Ok, this is my usecase:
I want to get full address by long;lat, with all administrative levels and districts.
With toponym HOUSE I get only first admin level, with street and house.
With toponym DISTRICT I get all admin levels and districts, but, of course, without street and house.

Then I combine them in full address.
Now I have to implement my own provider.

PR #469, opened 2015-11-05
Fixes #468

### LINKED ISSUE #468: Wrong order of sprintf`s params in Geocoder\Model\AdminLevelCollection::checkLevel
Now it is: 
throw new \OutOfBoundsException(sprintf(
   self::MAX_LEVEL_DEPTH,
   "Administrative level should be an integer in [1,%d], %d given",
   $level
));
//Output: 5

It should be:

throw new \OutOfBoundsException(sprintf(
   "Administrative level should be an integer in [1,%d], %d given",
   self::MAX_LEVEL_DEPTH,
   $level
));
//Output: Administrative level should be an integer in [1,5], 6 given

bubnovKelnik:
Fix wrong order of sprintf`s params in Geocoder\Model\AdminLevelCollection::checkLevel


willdurand:
thanks!

PR #471, opened 2015-11-30
ignore tests for export

laufhannes:
Please ignore your test suite for export, otherwise it will blow the vendor directory due to that large `GeoLiteCity.dat`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+.gitattributes export-ignore
+phpunit.xml.dist export-ignore
+tests export-ignore


willdurand:
Are you sure it fixes your issue with the large `.dat` file?

laufhannes:
Yes, just tested it with

```
{
	"repositories":
	[
		{
			"type": "vcs",
			"url": "http://github.com/laufhannes/Geocoder"
		}
	],
    "require": {
        "willdurand/geocoder": "dev-patch-gitattributes"
    }
}
```

and `composer install --prefer-dist`. This does not load any files marked with `export-ignore`.

willdurand:
:+1:

PR #474, opened 2015-12-05
Adds coutry-code to MapQuest-Results

heiglandreas:
The MapQuest API returns a two-letter country-code and no country name. Currently that Country-code is returned as country name when using the MapQuest-Adapter. But it is not returned as CountryCode though.

This PR solves that by adding the CountryCode.

To handle BC returning that 2 letter code as country-name is kept AS IS.


willdurand:
thanks!

Issue #470, opened 2015-11-06
Tests are broken with PHP 7


willdurand:

philippfrenzel:
Hi, it looks like google changed something with their geocode api landing point - so curl requests fail... I guess this crashes your tests... I didn't figure out a work around, would be great if someone would post a solution to this? Thanks!

Baachi:
@philippfrenzel Are you sure? Our tests only failing on PHP7 on all other platforms the tests passes.

philippfrenzel:
Hey, it's a other issue I'm facing. I make the requests "out of a docker container" and it looks like the routing is wrong... sorry for messing this up! Thanks for taking care!

mattsches:
Has anyone looked into this yet? If I run the tests locally against a freshly compiled PHP 7, I'm not getting any errors.

PR #475, opened 2015-12-05
Fixes Travis-Build

heiglandreas:
This adds a check for PHP7 to the HHVM-Check. Therefore on PHP7 the
PECL-GeoIP-Extension shouldn't be loaded. Instead the
geoip/geoip-package is required on PHP7.

It could also be required for HHVM, but I'm not sure that works...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,10 @@ matrix:
           env: deps="low"
 
 before_script:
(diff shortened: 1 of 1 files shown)


mattsches:
Fixes #470

willdurand:
awesome, thank you very much!

PR #478, opened 2015-12-16
Added postal_town to switch/case for locality address components

watchinharrison:
The locality for most UK addresses is registered under the type postal_town.


.gitignore
```diff
@@ -3,3 +3,4 @@ composer.lock
 composer.phar
 phpunit.xml
 php-cs-fixer.phar
+.idea
```
willdurand: would be nice not to change this file :)

willdurand:
Hi! Thanks for contributing. Sounds good to me, however would you mind add a test case?

watchinharrison:
By test case, do you mean a return response from the google geocode api that includes the type postal_town? 

http://maps.googleapis.com/maps/api/geocode/json?address=CF37,%20United%20Kingdom
http://maps.googleapis.com/maps/api/geocode/json?address=SR6,%20United%20Kingdom
http://maps.googleapis.com/maps/api/geocode/json?address=EH33,%20United%20Kingdom

are a few.

willdurand:
@watchinharrison I rather meant this: https://github.com/geocoder-php/Geocoder/commit/47baf0c5a707f9324f4bcbc5e623308160274685#diff-f97d2f35dbaa1150b96b68f1e1869885R348, but I wrote it for you that time ;-)

Thanks you for your patch, it has been added to the `master` branch (cf. 47baf0c5a707f9324f4bcbc5e623308160274685).

PR #481, opened 2015-12-31
Tom Lous

TomLous:
- Added accuracy & match properties in address 
- Added new Provider (https://geocode.farm/)


willdurand:
The test suite should be updated first.

Issue #480, opened 2015-12-23
HTTPLug


Baachi:
Hey guys,

what are youre thoughts about httplug? It's an http layer which follows the psr7 context and is
similar to ivory http adapter. Its well supported and its support async http calls.

If everyone would agree, i would start to integrate it in next couple of days.

@geocoder-php/geocoder

willdurand:
This is a must have to me. Yet we will have to release Geocoder **v4** (which is perfectly fine).

willdurand:
See #512

PR #487, opened 2016-01-18
Implement httplug

### LINKED ISSUE #484: "egeloen/http-adapter" deprecated
Composer message:

> Package egeloen/http-adapter is abandoned, you should avoid using it. Use php-http/httplug instead.

You should remove "egeloen/http-adapter" from require, and instead move it to `require-dev`.
This suffices for testing but allows people to use other not deprecated packages instead.
In general it is always a good idea to not directly include those proposed dependencies in require, but keep them in require-dev or suggest.

Baachi:
I started to integrate HTTPlug into geocoder.
Current status: Its integrated but i'm not about some implentation details. For example i don't know if we really need the `php-http/discovery` package because it 
introduce a lot of packages for a small benefit.

Reviews and thoughts are welcome!

Fixes #484 and #480


composer.json
```diff
@@ -26,6 +30,7 @@
         "geoip2/geoip2": "If you are going to use the GeoIP2DatabaseProvider.",
         "symfony/stopwatch": "If you want to use the TimedGeocoder"
     },
+    "minimum-stability": "beta",
```
joelwurtz: It can be removed, httplug is now in a stable version :)

composer.json
```diff
@@ -13,12 +13,16 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "egeloen/http-adapter": "~0.8|~1.0",
-        "igorw/get-in": "~1.0"
+        "igorw/get-in": "~1.0",
+        "psr/http-message-implementation": "^1.0",
+        "php-http/client-implementation": "^1.0",
+        "php-http/message-factory": "^1.0"
```
joelwurtz: I think it miss the require on php-httplug, even if it's included by the client implementation, its the direct package that it is linked too.

composer.json
```diff
@@ -13,12 +13,16 @@
...
     },
     "require-dev": {
         "geoip2/geoip2": "~2.0",
-        "symfony/stopwatch": "~2.5"
+        "symfony/stopwatch": "~2.5",
+        "php-http/discovery": "^0.6.4",
```
joelwurtz: Use discovery on 0.7 a lot of dependencies has been removed

joelwurtz:
You should ignore the `.puli` directory

src/Geocoder/Provider/AbstractHttpProvider.php
```diff
@@ -18,27 +21,28 @@
...
+     * @param HttpClient $client An HTTP adapter
      */
-    public function __construct(HttpAdapterInterface $adapter)
+    public function __construct(HttpClient $client = null, MessageFactory $factory = null)
     {
         parent::__construct();
 
-        $this->adapter = $adapter;
+        $this->client = $client ?: HttpClientDiscovery::find();
+        $this->factory = $factory ?: MessageFactoryDiscovery::find();
```
joelwurtz: One of the way we have done this for not having discovery, it to put discovery in suggest package then on the constructor we check the existence of the DiscoveryClass, if it has not been found we throw a \LogicException with a specific error which will be in our documentation for easy googling on that, here is an example : 

```php
if (null === $responseFactory && !class_exists('\Http\Discovery\MessageFactoryDiscovery')) {
    throw new \LogicException('No response factory provided and no discovery service is present to guess it, maybe you need to install php-http/discovery package?');
}
```

joelwurtz:
Also `CachedResponseClient` can maybe be remove and use the PluginClient from php-http which provides a CachePlugin working with a PSR6 implementation.

composer.json
Baachi: Sadly no, because the discovery depends on puli which is in beta state currently :cry:

Baachi:
It seems puli have a problem with hhvm and php7, because travis passes for php 5.5, 5.6 but not for php7 and hhvm.

src/Geocoder/Provider/AbstractHttpProvider.php
Baachi: I'm not sure about the discovery service at all, it brings many problems and puli as new dependency. Maybe we should wait for a stable release of the discovery service.

Baachi:
Some notes: Currently the tests don't passes on PHP7 and HHVM because puli is sadly not compatible with that. 
Also we must bump the requirement to php 5.5 because httplug needs at least php 5.5.

composer.json
```diff
@@ -13,12 +13,23 @@
...
+        "php-http/httplug": "^1.0"
     },
     "require-dev": {
         "geoip2/geoip2": "~2.0",
-        "symfony/stopwatch": "~2.5"
+        "symfony/stopwatch": "~2.5",
+        "php-http/mock-client": "^0.2",
+        "php-http/message": "^1.0",
+        "php-http/guzzle6-adapter": "^1.0",
+        "php-http/discovery": "^0.7.0",
```
Nyholm: Use "php-http/discovery": "^0.8.0" and put that in require.

src/Geocoder/Provider/AbstractHttpProvider.php
```diff
@@ -18,27 +21,33 @@
...
     }
 
     /**
      * Returns the HTTP adapter.
      *
-     * @return HttpAdapterInterface
+     * @return HttpClient
      */
-    public function getAdapter()
+    public function getHttpClient()
```
Nyholm: Maybe add public setters and protected setters. Otherwise we have to modify the constructors of every provider.

src/Geocoder/Provider/AbstractHttpProvider.php
```diff
@@ -18,27 +21,33 @@
...
+    /**
+     * @param HttpClient $client An HTTP adapter
+     */
+    public function __construct(HttpClient $client = null, MessageFactory $factory = null)
     {
         parent::__construct();
 
-        $this->adapter = $adapter;
+        $this->client = $client ?: HttpClientDiscovery::find();
+        $this->messageFactory = $factory ?: MessageFactoryDiscovery::find();
```
Nyholm: Put these in the getters. This will allow us not to use puli discovery by using `$provider->setHttpMessageFactory($factory)`

src/Geocoder/Provider/ArcGISOnline.php
```diff
@@ -167,8 +167,9 @@ private function buildQuery($query)
      */
     private function executeQuery($query)
     {
-        $query   = $this->buildQuery($query);
-        $content = (string) $this->getAdapter()->get($query)->getBody();
+        $query = $this->buildQuery($query);
+        $request = $this->messageFactory->createRequest('GET', $query);
```
Nyholm: Use $this->getHttpMessageFactory() here and on other providers.

smartpierre:
any updates ?

Nyholm:
If there is interest from the maintainers to merge this, I will be happy to complete this PR.

willdurand:
Yes, this should be merged into the lib at some point, especially because the current http lib is deprecated now.

cc @egeloen

Baachi:
Feel free to continue the work.

willdurand:
See #512

PR #492, opened 2016-02-02
Add throws to geocode/reverse methods of Geocoder interface.

### LINKED ISSUE #491: Include @throws in Geocoder interace geocode/reverse method docblock
If the recommended way of determining if there were errors during geocoding is to catch exceptions throw by the providers inside the Geocoder it would make reading the code more instructive to include an @throws annotation on the geocode/reverse methods and perhaps a @see method indicating where to look to understand how errors are handled

benglass:
Closes #491


willdurand:
thank you!

Issue #369, opened 2014-12-03
Googlemaps not always returning a bounds, use viewport?


zhuston:
I recently discovered that Googlemaps doesn't always return a bounds but when it doesn't it returns a viewport. What would you think about the Googlemaps provider returning the viewport if the bounds does not exist? 

In my use case it would be helpful as the viewport is better than nothing but I don't have a way to access the returned viewport with this library. Or if a way was added so I could access the viewport, unless it already exists and I am missing it.

Google docs, about 1/3 of the way down outlines Results format with "bounds" specified as optionally returned.
https://developers.google.com/maps/documentation/geocoding/

I am not sure what causes bounds not to be returned. Here are a couple of example zip code searches that do not return a bounds.
28299, 30149 or 30061
https://maps.googleapis.com/maps/api/geocode/json?address=28299

PR #496, opened 2016-02-18
Use viewport as fallback

belendel:
Hi,

I found issue #369 but noticed it was closed. Need the functionality in my application so decided to make this PR. It will use the viewport as fallback when bounds are not present.

### DIFF
--- a/src/Geocoder/Provider/GoogleMaps.php
+++ b/src/Geocoder/Provider/GoogleMaps.php
@@ -197,6 +197,13 @@ class GoogleMaps extends AbstractHttpProvider implements LocaleAwareProvider
                     'north' => $result->geometry->bounds->northeast->lat,
                     'east'  => $result->geometry->bounds->northeast->lng
                 );
+            } elseif (isset($result->geometry->viewport)) {
+                $resultSet['bounds'] = array(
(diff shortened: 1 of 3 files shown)


p-j:
I've impleted it a bit differently : https://github.com/vizeat/Geocoder/commit/be62d3bacaca73295cc4ff69ed5b6037726936c7

I had a specific need : make sure I get viewport whenever possible.
A quick explanation why : 
https://www.google.com/maps/d/u/0/edit?mid=zily8X7VKwgI.kgG-JKBOOxZs
In red you have the result of the `bounds` respectively for UK & Chile.
In yellow and green you have the results for the `viewport`

I needed to get bounds that better fits the expected result for the end user rather than the strictly correct result.

I haven't written tests for it so I'm not going to make a pull request but I have its running in production for us at the moment.

Nyholm:
I like the current implementation. But I agree with @p-j, your implementation gives better results... but it does not relate to the issue. 

@belendel I would like to merge this one but the tests fails. Would you mind rebasing on master and fix the tests?

Nyholm:
Ping. Can you rebase this PR and make sure the tests passes?

Nyholm:
Closed by #689

Nyholm:
Closed by #680

PR #501, opened 2016-03-01
Add support for setting bounds on the request with GoogleMaps

leewillis77:
The attached patch adds support for setting the bounds on GoogleMaps provider requests which allow you to set a preference for the results that are returned. See:

https://developers.google.com/maps/documentation/geocoding/intro#geocoding

### DIFF
--- a/src/Geocoder/Provider/GoogleMaps.php
+++ b/src/Geocoder/Provider/GoogleMaps.php
@@ -50,6 +50,11 @@ class GoogleMaps extends AbstractHttpProvider implements LocaleAwareProvider
     private $apiKey;
 
     /**
+     * @var string
+     */
+    private $bounds = null;
+
+    /**
      * @param HttpAdapterInterface $adapter An HTTP adapter
      * @param string               $locale  A locale (optional)
      * @param string               $region  Region biasing (optional)
@@ -108,6 +113,13 @@ class GoogleMaps extends AbstractHttpProvider implements LocaleAwareProvider
         return $this;
     }
 
+    public function setBounds($bounds)
+    {
+        $this->bounds = $bounds;
+
+        return $this;
+    }
+
     /**
      * @param string $query
      *
@@ -125,6 +137,10 @@ class GoogleMaps extends AbstractHttpProvider implements LocaleAwareProvider
 
         if (null !== $this->apiKey) {
             $query = sprintf('%s&key=%s', $query, $this->apiKey);
+        }
+
+        if (null !== $this->bounds) {
+            $query = sprintf('%s&bounds=%s', $query, $this->bounds);
         }
 
         return $query;


willdurand:
This would break the "common provider API".. @geocoder-php/geocoder WDYT?

vicchi:
Almost all of the "conventional" street level geocoding services support constaining results via a bounding box. A quick dive through various provider's documentation shows this is supported by ArcGIS, Google, Mapzen, Nominatim, TomTom and Yandex.

So this isn't an unreasonable enhancement IMHO.

We're also coming up to thinking about what Geocoder 4.0 will look like and as we're hiking the major version number, this is a good time to think about expanding the current common provider API.

This will need some thought about how to represent this consistently through the various providers as well as how to handle cases where this isn't supported.

As a concept, I think this is a valid one, but not the way in which it's implemented in this PR though.

stevevance:
+1 to having the bounding box parameter. This can drastically improve search results.

Baachi:
I'm not a big fan of it, as @willdurand said it will break the "common provider API". If you use this feature, you can't change the provider easily.

Nyholm:
I agree with both @willdurand and @Baachi. This library is meant to give a "common provider API". That means that a third party library author should be able to write something like this and be confident that the code works no matter what geocoder is provided. 

```php
class ThirdPartyCode
  public function __construct(Geocoder $gecoder) {
    $this->geocoder = $geocoder;
  }
  public function bar() {
    $result = $this->geocoder->geocode('foobar');
    // ...
  }
```

The third party library does not care if you change the geocoder provider. This is called the Liskov substitution principle. 


--------------

**However**. A application author should be able to make a decision that he/she never will change the provider. The application code may not type hint for `Geocoder` but for `GoogleMaps` or `GeocoderThatSupportsBoundsInterface`. 

```php
class ApplicationCode
  public function __construct(GoogleMaps $gecoder) {
    $this->geocoder = $geocoder;
  }
  public function bar() {
    $this->geocoder->setBounds(/* .. */);
    $result = $this->geocoder->geocode('foobar');
    // ...
  }
```

---------------

My point is that this library should define a *contract* which all geocoder providers MUST follow. But a specific provider (or a group of providers) may have some extra features. I see to issues with supporting that. I am 👍  for a PR like this.

src/Geocoder/Provider/GoogleMaps.php
```diff
@@ -108,6 +113,13 @@ public function setRegion($region)
         return $this;
     }
 
+    public function setBounds($bounds)
```
Nyholm: Shouldn't this parameter be type hinted with `Bounds`?

src/Geocoder/Provider/GoogleMaps.php
```diff
@@ -127,6 +139,10 @@ protected function buildQuery($query)
             $query = sprintf('%s&key=%s', $query, $this->apiKey);
         }
 
+        if (null !== $this->bounds) {
+            $query = sprintf('%s&bounds=%s', $query, $this->bounds);
```
Nyholm: This needs to be updated since `$this->bounds` should be a `Bounds` object.

Nyholm:
I just created https://github.com/geocoder-php/Geocoder/pull/576

Nyholm:
This is now supported with `GeocodeQuery::setBounds`

Nyholm:
Sorry, it was only the Query that had bounds support. I added a PR to make sure you can use bounds with google. See https://github.com/geocoder-php/Geocoder/pull/627

PR #502, opened 2016-03-06
Update PHP version syntax

Nyholm:
We should be more specific with the PHP version. We are not testing for PHP6, PHP8 and PHP9 but our composer file said that we support it. =)


willdurand:
thanks!

Nyholm:
Thank you for merging.

PR #503, opened 2016-03-10
fix: use statement for a phpdoc return tag

mocrates:
I don't know if version 2.x  is still maintained ... but if it can be help.


willdurand:
thanks!

PR #504, opened 2016-03-11
added geo array dumper

norkunas:


willdurand:
LGTM 👍

willdurand:
cc @geocoder-php/geocoder

tests/Geocoder/Tests/Dumper/GeoArrayTest.php
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * @author Jan Sorgalla <jsorgalla@googlemail.com>
+ * @author William Durand <william.durand1@gmail.com>
+ */
+class GeoArrayTest extends TestCase
+{
+    private $dumper;
+
+    public function setUp()
```
Baachi: should be protected.

tests/Geocoder/Tests/Dumper/GeoArrayTest.php
```diff
@@ -0,0 +1,133 @@
...
+            'geometry' => array(
+                'type' => 'Point',
+                'coordinates' => array(0, 0)
+            ),
+            'properties' => null
+        );
+
+        $result = $this->dumper->dump($address);
+
+        $this->assertTrue(is_array($result));
```
Baachi: Should be `assertInternalType``

norkunas:
I wonder why this package doesn't include phpunit in require-dev? Hard to know which phpunit version is used for testing

willdurand:
Great! Thanks @norkunas. One more thing: could you please squash your commits?



> I wonder why this package doesn't include phpunit in require-dev? Hard to know which phpunit version is used for testing

I never considered this useful since, most of the time, we only use common testing stuff that work with all PHPUnit versions.

norkunas:
I'm not familiar with squashing :/ Tried to do "git rebase -i HEAD~2" and now got a merge

PR #505, opened 2016-03-27
Package egeloen/http-adapter is abandoned

Jaspur:
Package egeloen/http-adapter is abandoned, you should avoid using it. Use php-http/httplug instead.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     ],
     "require": {
         "php": "^5.4|^7.0",
-        "egeloen/http-adapter": "~0.8|~1.0",
+        "php-http/httplug": "~1.0",
         "igorw/get-in": "~1.0"
     },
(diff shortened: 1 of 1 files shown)


philippfrenzel:
do you know if egeloen is generally discontinued?

Jaspur:
@philippfrenzel uh yea, see his repo.

egeloen:
I'm a little bit late here but be aware egeloen/http-adapter is still maintained but only for bugfixes and mainly because it is compatible with PHP 5.3 and 5.4 whereas httplug (it's successor) is only compatible with PHP >= 5.5 (the latest maintained PHP version http://php.net/supported-versions.php)

For more informations, see #487 for a current WIP PR which will fix the issue.

willdurand:
Closing in favor of #512

PR #510, opened 2016-06-01
Shamyyl patch geonames population

shamyyl:


Baachi:
Is geonames the only provider, which provides this information?

shamyyl:
I don't know. maybe.

Baachi:
Then i can't accept this PR. We want to provide a unified API.

PR #512, opened 2016-06-12
Implement Httplug

Nyholm:
This PR will replace #487. It continues the work and includes only some minor fixes.


Baachi:
@Nyholm awesome :+1: Thank you very much :tada:

Nyholm:
I've updated the PR. I've also updated the docs. This PR is ready for the final review. 

Im using `php-http/discovery:0.9` which make discovery more lightweight and easier to use.

willdurand:
:+1:

Nyholm:
Thank you for the reviews. Is there any questions or suggestions on this PR? Any thing I can do to help getting this merged?

willdurand:
I am fine with the PR, thanks @Nyholm! I let @geocoder-php/geocoder review & merge it.

Nyholm:
Thank you. 

FYI. Within an hour or two we will tag 1.0 of php-http/discovery. There are no BC breaks. I will update this PR to make sure we have stable dependencies.

Nyholm:
HTTPlug is using PSR-7 but PSR-7 does not include a way of creating or sending HTTP Requests. HTTPlug is an abstraction above libraries sending HTTP messages so we do not need a dependency on Guzzle, Buzz or any other library. 

This will make sure we follow the dependency inversion principle. ( The D in SOLID )

shadowhand:
And would depending directly on Guzzle be a detriment? Has anyone actually requested the ability to use a different HTTP client?

Nyholm:
Yes, it would. Say that we were depending on Guzzle6. That would mean that everybody with applications using Guzzle5 could not use this library. It would also mean that we have to update our code when Guzzle7 is released. 

That is why we need an abstraction to reduce coupling to implementations.

Baachi:
Can be merged!

composer.json
```diff
@@ -12,13 +12,21 @@
...
+        "php": "^5.5 || ^7.0",
+        "igorw/get-in": "^1.0",
+        "psr/http-message-implementation": "^1.0",
+        "php-http/client-implementation": "^1.0",
+        "php-http/message-factory": "^1.0.2",
+        "php-http/httplug": "^1.0",
+        "php-http/discovery": "^1.0"
     },
     "require-dev": {
+        "phpunit/phpunit": "^4.8",
```
michaelcullum: Why not define a bin directory as `/bin` and avoid executing `/vendor/bin/phpunit`?

michaelcullum:
It's a curious debate as ultimately you are still depending on an implementation, just one of a higher level than the http client itself. But that level of abstraction does remove a dependency on one specific http client which might help improve usage in larger applications using other http clients already. It changes the dependency from 'Guzzle (An http client)' to 'HttpPlug and an http client (but we don't care which one)' and the http client is what is more likely to be swapped in so does help make the lib more interoperable on balance. :+1:

sagikazarmark:
Thanks @michaelcullum, you summarised perfectly what goal we are trying to achieve. Also many thanks to @Nyholm who does the most to actually achieve that goal.

@shadowhand I accept that you don't want to see an HTTP abstraction layer in your projects, but please stop trolling. Do you think your question is valid in a project which provided such layer from the very beginning?

Flaxis:
Would be cool if someone could merge this PR. I really need this in my application.

willdurand:
@Nyholm can you merge this PR?

src/Geocoder/Provider/ArcGISOnline.php
```diff
@@ -40,13 +40,13 @@ class ArcGISOnline extends AbstractHttpProvider implements Provider
     private $protocol;
 
     /**
-     * @param HttpAdapterInterface $adapter       An HTTP adapter
+     * @param HttpClient $client       An HTTP adapter
```
sagikazarmark: Parameters not aligned

tests/Geocoder/Tests/TestCase.php
```diff
@@ -27,19 +28,16 @@ protected function getMockAdapter($expects = null)
             ->method('__toString')
             ->will($this->returnValue(''));
 
-        $response = $this->getMock('Psr\Http\Message\MessageInterface');
+        $response = $this->getMock('Psr\Http\Message\ResponseInterface');
```
sagikazarmark: I would recommend using actual implementations for testing instead. This is what we are going to do in HTTPlug libraries as well. The reasoning behind is that HTTP Messages can be interpreted as ValueObjects/DTOs as well and as such, they shouldn't be mocked. The actual behaviour is not important, rather the value they carry.

tests/Geocoder/Tests/TestCase.php
```diff
@@ -54,30 +52,31 @@ protected function getMockAdapterReturns($returnValue)
...
      * Because I was bored to fix the test suite because of
      * a change in a third-party API...
      *
-     * @return HttpAdapterInterface
+     * @return HttpClient
      */
     protected function getAdapter($apiKey = null)
     {
-        return new CachedResponseAdapter(new CurlHttpAdapter(), $this->useCache(), $apiKey);
+        return new CachedResponseClient(
```
sagikazarmark: Is this something that could be replaced by VCR in the future?

Nyholm: Maybe, I'll have to look in to that.

Nyholm:
Thank you for the feedback. I will merge this when Travis had a change to run again.

willdurand:
:+1:

ABM-Dan:
@willdurand any chance this might be released soon as 3.3.1 or 3.4.0?

Nyholm:
This is a major BC break, so we need to release 4.0. 

Before we tag that release I suggest we should look over the code if there is any more BC breaking changes we should do before making a release. 

@ABM-Dan feel free to test dev-master in your application to see if you can find some bugs that were not discovered by any of us reviewing this.

ABM-Dan:
Unfortunately I use this as part of BazingaGeocoderBundle, and there's no branches of it that use the dev-master version, even tho they are supposed to be synchronized.

sagikazarmark:
You can use branches as aliases in your project:

``` yaml
"willdurand/geocoder": "dev-master as 3.1",
```

This will satisfy `~3.1`, although you might need to override some services manually.

ABM-Dan:
Hm, too risky, I guess I will wait for BazingaGeocoderBundle 5.

willdurand:
> Before we tag that release I suggest we should look over the code if there is any more BC breaking changes we should do before making a release.

:+1: (note that @Nyholm is an official maintainer of Geocoder)

> Hm, too risky, I guess I will wait for BazingaGeocoderBundle 5.

You might want to use a Git commit hash to avoid issues.

Flaxis:
Thanks for merging! I was already working in this branch through a forked repository. I haven't found any bugs (yet) using the nominatim endpoint.

Romain:
Hi guys,
I'm using your package, and I still see the warning about @egeloen package which is abandoned.
Do you think you could merge this fix anytime soon?
Thanks
Romain

egeloen:
@Romain This is because your composer.json file should still rely on stable version of this library but this commit has only been added to the master branch (not yet released). Two possibilites for you:

 * Keep the code as it is (deprecated does not mean buggy) and upgrade when the new major version is released.
 * Upgrade to the master branch (as your risk) and upgrade your code accordingly as there are probably BC breaks as it is a new major release.

ABM-Dan:
I guess this simply pushes the issue of "when will we see the next release?"

Nyholm:
Have a look here: https://github.com/geocoder-php/Geocoder/milestone/3

It is basically [this PR](https://github.com/geocoder-php/Geocoder/pull/529) (input are welcome) and some minors. 

Im going to lock this thread so we do not ping all participants now when we go off topic. Feel free to open new issues. =)

PR #518, opened 2016-08-01
Update composer.json to allow symfony 3 component usage

jameshalsall:


Nyholm:
This will not effect your application. `require-dev` is only used when this composer.json is in the root. Ie on Travis.

jameshalsall:
Ahh that's awesome, thanks.

PR #520, opened 2016-08-18
Bug with the country name in MapQuest

artttj:
It has been used the same keys for country name and country code, so there was two-letter country code instead of the name.

### DIFF
--- a/src/Geocoder/Provider/MapQuest.php
+++ b/src/Geocoder/Provider/MapQuest.php
@@ -159,7 +159,7 @@ class MapQuest extends AbstractHttpProvider implements Provider
                     'locality'    => $location['adminArea5'] ?: null,
                     'postalCode'  => $location['postalCode'] ?: null,
                     'adminLevels' => $admins,
-                    'country'     => $location['adminArea1'] ?: null,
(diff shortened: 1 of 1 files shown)


Nyholm:
Im not too sure about this. Looking at the [documentation](https://developer.mapquest.com/documentation/geocoding-api/address/get/) the key `adminArea2` is not documented.

artttj:
@Nyholm it seems like some kind of undocumented feature. But it works.
You could see that it used in JS API reference.
https://developer.mapquest.com/documentation/javascript-api/geocoding/

Anyway, the two-letter code is not actually a country name. And that part of the code is just a duplicate.

Nyholm:
I see that they are used in the JS example but the output is blank if I understand it correctly. 

You are probably correct, it may work in most situations. I just wanted to highlight that this is not a no-brainer to merge this because it is not documented. 

Have you asked the mapquest.com people about this and why it isn't in the docs?

artttj:
I have looked through their forum and found this: https://developer.mapquest.com/forum/full-country-name.
Yeah, seems like there is no opportunity to get the full country name for now.
Closed.

PR #521, opened 2016-08-18
Adding cookbook

Nyholm:
Im not sure if we want this in the readme since it it out of scope. But is (IMHO) a really common use case and it would be nice for our users not having to figure this out themselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -527,6 +527,7 @@ Cookbook
 We have a small cookbook where you can find examples on common use cases: 
 
 * [Caching responses](/docs/cookbook/cache.md)
+* [Configuring the HTTP client](/docs/cookbook/http-client.md)
 
 
 Contributing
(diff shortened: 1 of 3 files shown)


willdurand:
It makes sense to have this in the doc. Maybe we could have a cookbook section or a FAQ? So that it clearlly indicates that this is out of this library's scope. What?

Nyholm:
I've started a new cookbook. =)

Nyholm:
I changed the title of this PR. I've added a cookbook with two entries: 
* how to use cache
* how to configure the HTTP client

These are normal uses but they are not in the scope for this library.

stevevance:
@Nyholm I already used your cache cookbook after stumbling upon it in the issues. Thanks for that!

Nyholm:
Im happy you got use of it!

Im just waiting for someone in the Geocoder team to review and merge.

willdurand:
🎉

Nyholm:
Thank you for merging.

PR #522, opened 2016-08-19
Formatted address

n-sviridenko:
Hello!

I'm understanding that you declined this feature. But despite the fact that this parameter isn't common, it's very useful for the localization, because we don't have a good php library for address localization yet. I'm sure that most of people are using google maps provider and also they're using this parameter. And it will be better to use your solution instead of writing own just for one property.

If you are not agree, would you please to add a raw response data of the providers?

Thank you for avance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ objects (`AddressCollection`), each providing the following API:
   properties);
 * `getCountryCode()` will return the ISO `country` code;
 * `getTimezone()` will return the `timezone`.
+* `getFormattedAddress()` will return the `formattedAddress`.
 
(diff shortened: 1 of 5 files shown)


n-sviridenko:
Also, you can create a compatibility table for each provider and each ability. It will warn users.

Nyholm:
Hey @n-sviridenko 

There has been a lots of changes in master since you made this PR. There is now a custom Google response where this change should be added. Can you rebase this PR and make sure you add the changes to the GoogleAddress instead of the shared one?

This PR is no longer on hold.

Nyholm:
Closed by #655

PR #528, opened 2016-09-08
Added full namespace to exceptions

Nyholm:
This is just a minor thing. PHPStorm was complaining about this.


willdurand:
ah right, 👍

Nyholm:
Thank you for the review.

Issue #523, opened 2016-08-22
[Address] Allow to extend address collection


egeloen:
Hey!

I would like to discuss an issue before you release the 4.x. Basically, in [egeloen/google-map](https://github.com/egeloen/ivory-google-map), I have added my own provider which allows to get full informations from the Google geocoder (and not just a subset as it is done here for consistency between providers).

Anyway, since the `AddressCollection` is final, I don't know how I can respect the contract of `geocode` and `reverse` (returning an `AddressCollection`) and at the same time being able to add more behaviors to it. In the doc, I explicitely explains that I broke the contract of `geocode` and `reverse` on purpose but it would be nicer if I can respect it...

Maybe introduce an interface for an extension point?

Feedbacks really welcome :)

egeloen:
I see that @Nyholm gives me a :+1: :) So, if there are no objections, I will make a PR tonight in order to introduce an interface.

Nyholm:
I gave a 👍  because I like the timing of this issue and because I like extensibility. I do believe in creating an interface because it will make us more SOLID.

willdurand:
Yep, having an extension point to expose more data or, at least, access raw data from providers, would close a lot of issues...

egeloen:
@willdurand AN interface would allow me to respect the contract of the main interface (returning an `AddressCollection`) whereas for now I can only use the `AddressCollection` since it's final.

Now, introducing an interface would just allow me to respect the main contract but I will still have to re-implement the `AddressCollection` & `Adress` logic by my own...

Can you clarify why do you make the result objects final?

egeloen:
I'm also :+1: for your idea about adding a new methods returning extra data. Then, I can use it instead of trying to extend the `Address*` models.

willdurand:
> Can you clarify why do you make the result objects final?

The initial idea was to provide some sort of value objects, only because I believe(d?) in a **common** layer for all the providers. That has always been the purpose of this lib, so that you could switch to any provider and you would end up with pretty much the same data. I know that it does not allow extension but it was intended. That's the "contract" of the lib.

**Yet**, more and more users want to extend the lib to get extra data. While I believe they should not use the lib for that, I also have to admit that Geocoder became the _de facto_ Geocoding lib in PHP. Hence, Geocoder v4 should be more open by providing extension points *or* a way to expose per provider extra data.

[...]

PR #529, opened 2016-09-08
Creating interfaces to be more SOLID

Nyholm:
This PR relates to #523 

To be more SOLID our value objects should have interfaces. This will allow others to write providers that may return more data than we have defined in our models. 

This is WIP because Im not happy with the *Interface suffix. We need to come up with better names. Or, are we happy with names like `AddressInterface`?

This PR also removes the `toString` function in favor for `__toString`
### TODO
- [x] Be confident with the interface names
- [x] Update changelog


willdurand:
To me, I would not add all these interfaces, because there are not different implementations for most of them (example the `BoundInterface` but also the collection one). If we provide a way to add user/custom data in an address object (from a provider), then there is no need for extra interfaces, because no one will change that.

I get that interfaces are often good for coupling, but there I don't really see the use case honestly (and I like [the SOLID stuff](http://williamdurand.fr/2013/07/30/from-stupid-to-solid-code/)).

unkind:
> To be more SOLID our value objects should have interfaces

Usually, interfaces for VOs is a sign of bad design.

Value objects are self-sufficient and very definite from behaviour point of view. There is no place for multiple behaviours here.

> This will allow others to write providers that may return more data than we have defined in our models.

It makes client-code coupled with specific implementation and it defeats the original purpose of the abstraction.

It makes sense to ask what exactly @egeloen wants to see in the `AddressCollection`.

Nyholm:
Thank you for valid input.. I've continued the work on this.. Will finish this week.

Nyholm:
I'm done with this PR now and would like to have feedback. 

The major changes I've done is to introduce a `GeocoderResult` and a `Position` interface. Those are implemented by the `AddressCollection` and `Address` object respectively.

I've also make sure that the `Position::getCoordinates`, `Position::getCountry` and  `Position::getBounds` always returns an object. These objects have a `isDefined` function. 

I think it is a good idea to remove `Position::getLatitude`, `Position::getLongitude` and  `Position::getCountryCode` because they are redundant. 


### Future work: 

One could add vendor specific `Positions` ie `GooglePosition` to include some extra data just available for Google. (Suggested by @unkind here: https://github.com/geocoder-php/Geocoder/issues/523#issuecomment-252421822)

unkind:
> One could add vendor specific Positions ie GooglePosition to include some extra data just available for Google

I'm not sure I understand how it is supposed to work. Interface is limited to specific method list, there is no way to get an extra data. `if ($position instanceof GooglePosition)`? Well, at least, it should not be called  "to be more SOLID", but "we make this hack for practical purposes". :)

> Suggested by @unkind here: #523 (comment)

I suggested to make completely independent interfaces, e.g. 

```php
interface GoogleGeocoder
{
    /**
     * @return GoogleAddress[]
     */
    public function geocode(...);
}
```

I mean, if you really want to provide vendor-specific address objects, you would have almost same amount of work in terms of lines of code (at least, I think so), but you'd have clean API for every major providers.

> I've also make sure that the Position::getCoordinates, Position::getCountry and Position::getBounds always returns an object. These objects have a isDefined function.

Offtopic: `isDefined` looks nasty, it's like having `$datetime->isValid()`. You either have valid object (with valid lat/lng) or you should not have it at all. Invalid object doesn't make sense.

unkind:
To make clear my point:

```php
final class GoogleGeocoder
{
    public function __construct(bool $useSsl, ...)
    {

    }

    public function geocode(string $address): GoogleAddress
    {
        // ...
    }

    public function geocodeWithin(string $address, BoundingBox $bbox): GoogleAddress
    {
        // ...
    }
}

final GoogleMaps implements Provider
{
    public function __construct(GoogleGeocoder $googleGeocoder)
    {
        // ...
    }

    public function geocode()
    {

    }
}
```

Users will configure the `GoogleGeocoder` and use it directly if they need some specific Google-specific methods/data; `GoogleMaps` will be simple adapter for generic `Geocoder`, it works as usual.

Nyholm:
> I'm not sure I understand how it is supposed to work. Interface is limited to specific method list, there is no way to get an extra data. 

The idea is that you can always be sure that `Geocoder::geocode` will give you a `GeocoderResult` with `Positions`. That is a contract all geocoders MUST follow. There is nothing that forbids the `GoogleGeocoder::geocode` to return an object like: 

```php
class GoogleGeocoderResult implements GeocoderResult {
  // ... All method in the GeocoderResult interface

  public function getQueryTime() {
    // Return the time it took to query the API
  }
}
```
And the `Position` object returned from GoogleGeocoder could also include more methods in the same way. 

> if ($position instanceof GooglePosition)? 

Yes, you would have to do `if ($position instanceof GooglePosition)`. But introducing interfaces like this would allow us to create other vendor specific interfaces. 
 
```php
interface GoogleGeocoder implements Geocoder
{
    /**
     * @return GoogleGeocoderResult[]
     */
    public function geocode(...);
}

interface GoogleGeocoderResult implements GeocoderResult
{
    // ...

   public function getQueryTime();
}
```

If an application author decides that he is only interested in using the Google geocoder he/she can type hint directly at the `GoogleGeocoder` interface. and no `if ($position instanceof GooglePosition)` is needed. At the same time third party libraries can use the same geocoder because it will still respect the `Geocoder` interface. 


> Well, at least, it should not be called "to be more SOLID", but "we make this hack for practical purposes". :)

I can rename the issue title if you want =)

--------------

> Offtopic: isDefined looks nasty, it's like having $datetime->isValid(). You either have valid object (with valid lat/lng) or you should not have it at all. Invalid object doesn't make sense.

Im not happy with that function either. It did already exist in the `Bounds` object.

unkind:
> There is nothing that forbids `GoogleGeocoder::geocode` to return an object like

Sure, but you open this knowledge to users like new kind of API, you legalize the magic with `instanceof`.  

> But introducing interfaces like this would allow us to create other vendor specific interfaces.

I wouldn't try to make inheritance tree like this. In practice, it becomes harder to understand what's going on and reduces flexibility: you can't change contracts of different providers independently. You lock your provider with method `getCountry: country { code, name }`, but what if some provider is allowed to provide more info of a country? In my example, I'd have `GoogleMaps/Country` value object with all possible country's data. Sure, you can add just `getExtraCountry` method which returns `GoogleMaps/Country` or something like this, but it hurts API. Just think about `GoogleGeocoder` interface like you want to build it from scratch. Do you need to inherit `Geocoder` to limit yourself? I don't think so.

Honestly, I'm not user of the library, so it's not fair for me to judge it. I can find reasons why it's not good in theory, but your approach may be way more practical.

Nyholm:
I like having this discussion with you. Thank you for sharing your thoughts. 

The purpose of this library is to have a common contract for the geocoded result. If one feel that geocoder X is giving me poor results one should easily be able to switch to geocoder Y. (This is the Liskov substitution principle.) This is something we strongly believe in and should not be up for change. This contract is enforced by `GeocodedResult` interface (currently with `AddressCollection` object). 

The reason I introduce the interfaces is that I want to open up the possibility for others to create a more specific result. If the application author feels it is important, he/she can "trade" Liskovs substitution principle to have this specific result. This will increase flexibility of the library. 

At the same time, he/she can still use the `AwesomeGeocoderResultLogger::log(GeocodedResult $result)` and all other possible third party libraries and/or plugins. 


> I wouldn't try to make inheritance tree like this. In practice, it becomes harder to understand what's going on and reduces flexibility.

The inheritance if only for interfaces. And yes, it reduces flexibility because we want to follow the `GeocodedResult`. 

Im not sure **we** want to introduce this extra `GoogleResult` interface. But other packages might. (#523)

unkind:
> The purpose of this library is to have a common contract for the geocoded result. If one feel that geocoder X is giving me poor results one should easily be able to switch to geocoder Y. (This is the Liskov substitution principle.)

I think you missed my point, I never suggested to return different objects to break LSP.

My point is if user wants to get more vendor-specific data/methods/abilities, he should use vendor-specific geocoder.

Let's say we want to query city's coordinates by city name within specific country. 

Instead of 

```php
final class UserlandService
{
    public function __construct(Geocoder $geocoder)
    {
        // ...
    }

    public function findTheBestResult(string $cityName, string $countryCode): Coordinates
    {
        // Order is not guaranteed.
        foreach ($this->geocoder->geocode($cityName) as $position) {
            if (
                $position instanceof GooglePosition
                && $position->getCoordinates()->isDefined()
                && $position->getCountryCode() === $countryCode
            ) {
                return $position->getCoordinates();
            }

            if ($position instanceof OpenStreetMapPosition
                && $position->getCoordinates()->isDefined()
                && $position->getCountryCode() === $countryCode) {
                return $position->getCoordinates();
            }
        }
    }
}
```

I prefer

```php
final class UserlandService
{
    public function __construct(GoogleGeocoder $google, OpenStreetMapGeocoder $osm)
    {
        // ...
    }

    public function findTheBestResult(string $cityName, string $countryCode): Coordinates
    {
        try {
            return $this->google->findFirstMatchWithCountry($cityName, $countryCode);
        } catch (NoResults $e) {
            foreach ($this->osm->geocode($cityName) as $address) {
                if ($address->getCountry()->getCode() === $countryCode) {
                    return $address->getCoordinates();
                }
            }
        }
        
        throw NoResults::cityWithinCountry($cityName, $countryCode);
    }
}
```

Yes, I have 2 dependencies, but it makes code more explicit and more efficient. The providers have different abilities, different API, different methods. Google can directly execute my query, but OSM does not. Also, I can make very optimized queries for my use cases if specific provider allows it. The generic Geocoder, on other hand, run all possible providers with the same query. It makes him less efficient for this kind of task. Different cases may require different tools.

Do you mind to provide the use case when the generic Geocoder would be better to work with vendor-specific stuff?

Nyholm:
> I think you missed my point

I think I did miss your point. I very much agree with your last post. That is my intention as well. That is exactly what Im trying to achieve with this PR. I do also want to allow something like this: 

```php
final class UserlandService {
  // ...
     $googleResult =  $this->google->geocode('foo');

     return $googleResult->getFirstResultThatAllowStreetView();
}
```

With the current code in master your preferred example and my example above is not possible. I **believe** that this PR makes it possible with the best way solution.

nicholasruunu:
I like these changes!

Possible improvements:
* `interface Position` feels a bit specific, I wouldn't necessarily connect an address to it for example. I feel like Location might be a better name.
* `getAdminLevels()` feels out of place in the Position interface, I'd try to extract that out.

unkind:
In the radical case you'd end up with a marker interface:

```php
interface Location
{
}

interface Geocoder
{
    public function geocode(...): Location;
}
```

It can be literally anything. Very flexible, so awesome.

Nyholm:
Thank you for your reviews. I agree and I've made updates accordingly. 

> getAdminLevels() feels out of place in the Position interface, I'd try to extract that out.

Sure. I know that one is weird. But a admin level is actually a fundamental property of a ~~Position~~ Location. It is an abstraction of getState, getMunicipality, getSwedishLän, getGermanBundesstadt etc etc. I do not think we can/should extract that outside of the Location.

Nyholm:
This looks like a quite big PR but it is a lot of documentation changes and changes in the tests. @geocoder-php/geocoder please look at the Changelog file to see if you agree with the changes. I want to come to a decision with this PR this week.

CHANGELOG.md
```diff
@@ -1,6 +1,20 @@
 CHANGELOG
 =========
 
+### 4.0.0 (2016-xx-xx)
+
+* Added: Interface for `Geocoder\Model\AddressCollection` called `Geocoder\GeocoderResult`. Public APIs are updated to type hint for `Geocoder\GeocoderResult`.
```
willdurand: I am not a fan of this to be honest. You repeat "geocoder" twice, and "result" does mean anything. I used `AddressCollection` exactly for that. It is meaningful and it is clear that we get a set of results. I don't see the point of changing this.

CHANGELOG.md
```diff
@@ -1,6 +1,20 @@
...
+* Added: Interface for `Geocoder\Model\Address` called `Geocoder\Location`. Public APIs are updated to type hint for `Geocoder\Location`.
+* Added: `Country::isDefined`
+* Added: `Cordinates::isDefined`
+* Changed: `Location::getCoordinates`, `Location::getBounds` and `Location::getCountry` will never return null.
```
willdurand: What is returned when there is no data for this then?

CHANGELOG.md
```diff
@@ -1,6 +1,20 @@
...
+* Removed: `AdminLevel::toString` in favor for `AdminLevel::__toString`.
+* Removed: `Country::toString` in favor for `Country::__toString`.
+* Removed: `Address::getCountryCode` in favor for `Address::getCountry()->getCode()`.
+* Removed: `Address::getLongitude` in favor for `Address::getCoordinates()->getLongitude()`.
+* Removed: `Address::getLatitude` in favor for `Address::getCoordinates()->getLatitude()`.
```
willdurand: Shortcuts are sometimes useful

CHANGELOG.md
Nyholm: I'm happy with any name but `AddressCollectionInterface`. Do you have any suggestions? Or do we want the "Interface" suffix?

CHANGELOG.md
Nyholm: They will always return a Coordinates/Bound/Country object. But `Bound::isDefined() == false`.

CHANGELOG.md
Nyholm: I agree that a *shorter way* would be useful. But do we really need both of these: 
* `Address::getCoordinates()->getLatitude()`
* `Address::getLatitude()`

I would like to drop one of them. I chose the latter because I liked the separation of coordinates into a separate object. But Im happy to change if you prefer.

CHANGELOG.md
willdurand: ok :+1:

CHANGELOG.md
Nyholm: Any suggestions? Input? @Baachi?

src/Geocoder/GeocoderResult.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function getIterator();
+
+    /**
+     * {@inheritDoc}
+     */
+    public function count();
```
Baachi: This methods can be removed right? The method is already covered by the interfaces.

CHANGELOG.md
Baachi: Why not just name it `Collection`? It is in the geocoder namespace, so it should be logical or not?

Nyholm: Im okey with that. :)

How about you, @willdurand?

willdurand: Ok with the name, but I still question the need for an interface here

Nyholm: Thank you for the feedback. I've updated the code

src/Geocoder/GeocoderResult.php
Nyholm: You are correct. Same with getIterator.

Nyholm:
I see that @unkind comment has been hidden in the diff. See https://github.com/geocoder-php/Geocoder/pull/529#discussion_r85295670

>Probably, it's better to deprecate existing isDefined and return null?
  
If we dropped the `isDefined` we have to make sure to validate that the argument parameters always have values. Ie: 

```php
final class Coordinates
{
    /**
     * @param double $latitude
     * @param double $longitude
     */
    public function __construct($latitude, $longitude)
    {
         if ($latitude === null || $longitude == null) {
            throw new Exception();
         }
        $this->latitude  = $latitude;
        $this->longitude = $longitude;
    }
}
```

Is that something we prefer over the isDefined functions?

CHANGELOG.md
Nyholm: > but I still question the need for an interface here

An interface wont hurt and it opens up the possibility for others custom providers to return a `CollectionOnSteroids` and still comply with the contract.

unkind:
> Is that something we prefer over the isDefined functions?

Moreover, I'd put assertions there:

```php
    public function __construct(float $latitude, float $longitude)
    {
        Assertion::between($latidude, -90, 90);
        Assertion::between($longitude, -180, 180);

        $this->latitude  = $latitude;
        $this->longitude = $longitude;
    }
```

Nyholm:
Yeah, I agree with you. Thanks!

CHANGELOG.md
Nyholm: As I see it, this is the only issue we have not agreed on. Do you have any feedback here?

unkind:
```
$googleResult->getFirstResultThatAllowStreetView()
```

How it is supposed to work? I mean, you can't make custom request here, because result is already here, right? Lazy loading magic?

Nyholm:
By introducing interfaces we **allow** ourselves and third party libraries to extend the functionality while **still keeping** the contract of a provider. 

`MyAwesomeGoogleProvider` may return an class like: 

```php
class SuperCollection implements \Geocoder\Collection
{
  // ... Implement methods in the Collection interface

  public function getFirstResultThatAllowStreetView(){
     /** @var AwesomeLocation $item */
     foreach ($this->items as $item) {
        if ($item->hasStreetView()) {
          return $item;
       }
     }
     return null;
  }
}

class AwesomeLocation implements \Geocoder\Location
{
    // ... Implement methods in the Location interface

    public function hasStreetView() {
       // ..
    }
}
```

unkind:
Yeah, I understand it.

However, one of my original points with different interfaces of Geocoder was
> Also, I can make very optimized queries for my use cases if specific provider allows it

It means that if Google, for example, provide `GET /search?street_view=true&limit=1` method, I can utilize it by calling `$googleGeocoder->geocodeWithStreeViewOnly(..., 1): GoogleResult`, but you suggest to use generic method `$googleGeocoder->geocode(...): Result` and **then** call custom one (`$googleResult->getFirstResultThatAllowStreetView()`). That makes me wonder: what does `Result` contain right after calling `geocode()`? Further, does `getFirstResultThatAllowStreetView` make HTTP-request?

Your value object (`GeocodingResult`/`AddressCollection`) becomes a service. It makes him way more complicated. Are you sure this approach is better than just different provider interfaces?

Nyholm:
> It means that if Google, for example, provide GET /search?street_view=true&limit=1 method, I can utilize it by calling $googleGeocoder->geocodeWithStreeViewOnly(..., 1): GoogleResult

You could add that method, no worries. =)

> Your value object (GeocodingResult/AddressCollection) becomes a service.

The interface defines a value object. Third party implementation may do whatever as long as they follow the contract by the interface. 

> Are you sure this approach is better than just different provider interfaces? 

One does not exclude the other. 

My point is, and has always been, that if you introduce interfaces you will increase flexibility for the providers (open for extension). At the same time the interfaces specifies a contract which makes providers respect the Liskov substitution principle. 

-------

I would be happy to discuss a PR where you introduce a `LocationWithStreetview` interface or an interface for providers that supports a `setBounds` function. 

But to create one interface for Google, one interface for Bing, one for ... etc is no good solution at all. Those interfaces are too specific and does not help interoperability.

unkind:
> My point is, and has always been, that if you introduce interfaces you will increase flexibility for the providers (open for extension)

The thing is you want to make `instanceof` an extension point in fact. I think it's a hack.

> But to create one interface for Google, one interface for Bing, one for ... etc is no good solution at all. 

You want to hide same methods behind the generic interface, but when I suggest to make them explicit, you call it "not good solution at all". Do I miss something? :)

> Those interfaces are too specific and does not help interoperability.

When one wants to use vendor-specific stuff, he doesn't need interoperability. 

You still can make generic and simple `Geocoder::geocode()` method, but add also vendor-specific geocoders:

```php
final GoogleGeocoder implements Geocoder
{
    public function geocode(): AddressCollection
    {
        // Usual generic contract
    }

    public function geocodeSpecific(): GoogleAddressCollection
    {
        // Same, but with extended methods
    }

    // More Google-specific methods
}

// or

final GoogleGeocoder 
{
    public function geocode(): GoogleAddressCollection
    {
        // ...
    }
} 

final GoogleGeocoderAdapter implements Geocoder
{
    public function geocode()
    {
        return $this->convertToGeneric($this->googleGeocoder->geocode(...));
    }
}
```

When one wants to use generic geocoder, he uses `Geocoder` contract; when someone wants to gain vendor-specific abilities, he may want to inject vendor-specific providers.

Nyholm:
> The thing is you want to make instanceof an extension point in fact. I think it's a hack.

No I do not. You type hint for what you need. Do you need a `Geocoder` or do you need a `GeocoderThatSupportsSetBounds`? You would never need both. Application authors knows what provider that is provided and third party libraries should use the `Geocoder` interface. 

> > But to create one interface for Google, one interface for Bing, one for ... etc is no good solution at all.

> You want to hide same methods behind the generic interface, but when I suggest to make them explicit, you call it "not good solution at all". Do I miss something? :)

I put the common things in the `Geocoder` interface. Nothing more because that would violate the interface segration principle. I agree with you that creating specified interfaces that cover 3-5 provider or a very specific use case are good. But to create 15 different interfaces for 15 different providers serves no purpose. 

Ie "vendor specific" are all wrong but "feature specific" are encouraged. 

> > Those interfaces are too specific and does not help interoperability.

> When one wants to use vendor-specific stuff, he doesn't need interoperability.
 
True and I very much agree. But in that case he doesn't need an interface either. =)

unkind:
> But in that case he doesn't need an interface either

Yes, as you can see my `GoogleGeocoder` is a concrete class, I'd inject it directly and type-hinted against it if I need Google-specific stuff. I said "interface" in a wider sense (class has its own interface).

> GeocoderThatSupportsSetBounds

Unfortunately, this class always will return the generic `Result` from interface point of view, you can't make

```
final class GeocoderThatSupportsSetBounds implements Geocoder
{
    public function geocode(): CustomAddressCollection
    {

    }
}
```

> But to create 15 different interfaces for 15 different providers serves no purpose.

Wow, don't create 15 different providers if you don't need them. Create specific providers for most popular vendors.

Nyholm:
> I said "interface" in a wider sense (class has its own interface).

Sure, my bad. I was thinking about actual interfaces when you said "interface". I did not consider "classes and interfaces". 

> Unfortunately, this class always will return the generic Result from interface point of view, you can't make

That is not really how you would use it... 

```php
interface GeocoderThatSupportsSetBounds {
  public function setBounds(Bounds $bounds);
}

final class MyGeocoder implements Geocoder, GeocoderThatSupportsSetBounds
{
    public function setBounds(Bounds $bounds) {

    }

    public function geocode(): Collection
    {
        
    }
}
```

You must always respect the `Geocoder` interface. And of course, MyGeocoder::geocode can always return something that extends the `Collection`.

> > But to create 15 different interfaces for 15 different providers serves no purpose.

> Wow, don't create 15 different providers if you don't need them. Create specific providers for most popular vendors.

We currently have 19 providers in this package and 10 more in the geocoder-extra package..

unkind:
> We currently have 19 providers in this package and 10 more in the geocoder-extra package..

I see. I didn't suggest to make 19 specific providers. As I said, only popular ones (in other words, those providers which you want to extend with specific result objects). The rest would just simply implement the only `geocode()` method as they already do.

> That is not really how you would use it...

I wouldn't suggest to make mutable services, honestly, i.e. `geocodeWithBounds(...)` is usually would be better than `->setBounds()->geocode()`.

> And of course, MyGeocoder::geocode can always return something that extends the Collection

And here `instanceof` takes a place? Or IDE hacks like `/** @var $result GoogleCollection */`? How would you use vendor-specific `Collection` methods?

Also, by replacing VOs with an interface, you make them harder to understand, to serialize, to replace-on-the-fly (`$address = new Address($address->getName(), ..., 'changed value')` <- you don't know original type of the VO).

Nyholm:
> I see. I didn't suggest to make 19 specific providers. As I said, only popular ones (in other words, those providers which you want to extend with specific result objects). The rest would just simply implement the only geocode() method as they already do.

Why? What is the win for that? 
I said that *feature* specific interfaces are good. But **why** vendor specific?

> I wouldn't suggest to make mutable services, honestly, i.e. geocodeWithBounds(...) is usually would be better than ->setBounds()->geocode().

This PR does not include such change. It just allows the possibility. Feel free to argue about the implementation details of my example in when that PR comes around. 

> And here instanceof takes a place?

Nope. 

```php
class MyGeocoder implements Geocoder{
  public function geocode(): MyCollection { //.. }
}
class MyCollection implements Collection {
  // ..
  public function extraFeature();
}

class Consumer {
  public function __construct(MyGeocoder $gecoder) {
    //
  }
}
```

> Also, by replacing VOs with an interface, you make them harder to understand, to serialize, to replace-on-the-fly ....

No no no. Do you mean that Guzzles Request object is harder to understand, serialize and "replace-on-the-fly" (?) because of PSR7? That is just ridiculous and it feels like you are reaching for arguments not to add 2 interfaces...

unkind:
> Why? What is the win for that? 

Clearness. Independence from each other. Yeah, sometimes some sort of duplication, but don't forget about [DRY tyranny](http://us3.campaign-archive2.com/?u=1090565ccff48ac602d0a84b4&id=5132867f6e).

> I said that feature specific interfaces are good. But why vendor specific?

Because people wants exactly it: they want to stick with specific provider(s) sometimes. They want to know origin of a data (they want explicitly say "first I need to check Google, then OSM"), they want to stick to google_location_id, they want to get station name from Yandex, etc.

> Nope.

You [can't](https://github.com/geocoder-php/Geocoder/pull/529#issuecomment-258986046) do that: https://3v4l.org/kv9pn
```
Fatal error: Declaration of MyGeocoder::geocode(): MyCollection must be compatible with Geocoder::geocode(): Collection in /in/kv9pn on line 16
```

> Do you mean that Guzzles Request object is harder to understand, serialize and "replace-on-the-fly" (?) because of PSR7? 

Not Guzzle Request by itself, but I think that PSR-7 is not example of great design, you are right. However, I don't think we can't discuss it here.

unkind:
To summarise, I see 3 options:

1. Your proposal.

    Advantanges:
    * no adapters needed;
    * it's possible to query the generic Geocoder (with `GeocoderAggregator`) and *then* get provider-specific stuff with `instanceof` from array of results (so-so advantage).

    Disadvantages:
    * additional interfaces for VOs;
    * hackish (in my opinion) way to deal with return type hint of `geocode()`.

    Example:

    ```php
    final class GoogleMapsResult implements Collection
    {
        // ...
    }

    final class GoogleMaps implements Geocoder
    {
        public function geocode(...): Collection
        {
            // This method also can return specific collection,
            // but in order to use it, you need to use instanceof.
        }

        public function geocodeWithExtendedGoogleSpecificStuff(...): GoogleMapsResult
        {
            // Google-specific stuff
        }
    }
    ```

2. My first proposal. 

    Advantages:

    * no interfaces for VO needed (true value objects without possible magic inside of them);
    * providers are not forced to be coupled with the generic Geocoder, you can move `GoogleProvider` to external package as a standalone solution for Google Maps API;

    Disadvantages:

    * more code to write (adapters) for providers with additional abilities.

    Example:

    ```php
    final class GoogleMapsResult
    {
        // There is no Collection interface to implement.
    }

    final class GoogleMapsProvider
    {
        public function geocodeLocation(...): GoogleMapsResult
        {
            // Google-specific stuff
        }
    }

    final class GoogleMapsAdapter implements Geocoder
    {
        public function __construct(GoogleMapsProvider $googleProvider)
        {
            // ...
        }

        public function geocode(...): AddressCollection
        {
            return $this->convertToGeneric($this->googleProvider->geocodeLocation(...));
        }
    }
    ```

3. My second proposal.

    Advantages:

    * no interfaces for VO needed (true value objects without possible magic inside of them);

    Disadvantages:

    * more code to write (adapters) for providers with additional abilities.
    * providers would have the generic `geocode()` method and you have to think how to name your provider-specific method (`geocodeWithGoogleStuff`? `geocodeExtended`?).

    Example:

    ```php
    final class GoogleMapsResult
    {
        // There is no Collection interface to implement.
    }

    final class GoogleMapsProvider implements Geocoder
    {
        public function geocode(...): AddressCollection
        {
            // ...
        }

        public function geocodeWithExtendedGoogleSpecificStuff(): GoogleMapsResult
        {
            // ...
        }
    }
    ```
    
---

As I can see, there is no big difference in our solutions (especially with my second proposal).

src/Geocoder/Location.php
```diff
@@ -0,0 +1,104 @@
+<?php
+namespace Geocoder;
+
+use Geocoder\Model\AdminLevelCollection;
+use Geocoder\Model\Bounds;
+use Geocoder\Model\Coordinates;
+use Geocoder\Model\Country;
+
+/**
+ * A position is a single result from a Geocoder.
```
willdurand: that is not a position but a location, right?

willdurand:
Hey! I reviewed both the comments and the changes.

I see interesting things in "both sides", but let me introduce something else: why not adding a "raw" attribute to the `GeocoderResult` (value) object? That way:

1. no much change in the current code base
2. no interfaces on VOs (which is kinda weird)
3. no interfaces/new methods for each provider
4. people get what they want: access to the underlying data

Sure we don't have a solid contract but, _heh_, that's not the purpose of the lib anyway. 90% of the issues related to this PR are: "I'd like to get this information from Google Maps".

Nyholm:
Thank you. 

I can agree with adding a `raw` function. But this does not stop us from introducing these 2 interfaces. 

I do aslo think I know ways of addressing the issues of 90% of those PRs. But this PR is a blocker for that.

unkind:
Some interesting discussion about interfaces for value objects: https://gist.github.com/Ocramius/a9d6713867929fe23f36b35981befb0d

willdurand:
If we go the "raw" way, then do we really need interfaces? Especially on VOs...?

Nyholm:
Interfaces are never a bad thing. And Im not sure these are actually value objects. From Wikipedia: 

> In computer science, a value object is a small object that represents a simple entity whose equality is not based on identity

This is not a "small object" nor a "simple entity". The *need* for a `raw` function is evidence of that.

willdurand:
The only interface would rather be `RawAware` something then.

Nyholm:
I would like to argue for making RawAware a complementary interface or a part of an existing one. In other words: I do not like to see that RawAware is our *only* interface.

unkind:
> Interfaces are never a bad thing

It's a strong statement. Interfaces can be sign of wrong abstraction. We don't need interface for `DateTime`, for example (note that existing `DateTimeInterface` is a hack, not a real interface and no one can implement it). We expect Gregorian `DateTime` and we don't expect something else. The behaviour must be the same, there is no room for different implementations, interface is harmful here.

> And Im not sure these are actually value objects

What exactly is not a VO? `AddressCollection`? It is a VO in the same sense as `HttpResponse` is a VO. It is response of the Geocoder. If you modify it, you'd have different response.

Nyholm:
@unkind I wrote the definition of a value object. Yes, interfaces **could** be a sign of wrong abstraction... But Im not inventing new interfaces, Im extracting interfaces from the models we got. 

I feel like you are not trying to move this PR to either be accepted or rejected. This discussion has been long as it is, there is no point of having a general (out of topic) discussion about interfaces in general. Im sorry if I have misunderstood or missinterpided your messages. 

About how one should use interfaces and different kinds of feature interfaces for the providers: https://3v4l.org/HdCVp

I want to accept or reject this PR. I think all arguments are on the table. After this PR I want to make sure to complete all tasks before we can tag a stable release.  I do not want to merge my own PR that's why I ask other maintainers to make the decision.

unkind:
> Interfaces are never a bad thing
> Yes, interfaces could be a sign of wrong abstraction...  But Im not inventing new interfaces, Im extracting interfaces from the models we got

So, basically, you say "we have model `Foo`. Extracting `FooInterface` is never a bad idea", right? So we have `DateTime`. Extracting `DateTimeInterface` is never a bad idea? I don't think so. 

 > I feel like you are not trying to move this PR to either be accepted or rejected

I'm definitely trying to show arguments against this PR. I like the idea about key-value `raw` attribute though.

willdurand:
Now let's continue the journey towards Geocoder 4.0 😉

Nyholm:
Thank you for merging

PR #530, opened 2016-09-12
Fix bug where cached unit test responses aren't serialised

vicchi:
Although the cached unit test responses are serialised, the unit test code framework doesn't actually serialise any new responses that are added to the cache directory. This fixes that.


willdurand:
Looks good to me, I think no one tried to rebuild the cached responses for a long time ;-)

Anyone else?

Nyholm:
Thank you!

PR #531, opened 2016-09-12
Add support for Mapzen's Geocoder

vicchi:
This PR depends on the serialisation fix in PR #530 in order for the new provider's unit test cache files to be written correctly and for the unit tests to be repeatable, based on those cached files.


Nyholm:
Awesome @vicchi!

Im 👍 but I do not know about the Mapzen service. I've reviewed this PR from the general providers' point of view.

src/Geocoder/Provider/Mapzen.php
```diff
@@ -0,0 +1,223 @@
...
+use Geocoder\Exception\InvalidCredentials;
+use Geocoder\Exception\QuotaExceeded;
+use Geocoder\Exception\NoResult;
+use Geocoder\Exception\UnsupportedOperation;
+use Http\Client\HttpClient;
+
+/**
+ * @author Gary Gale <gary@vicchi.org>
+ */
+class Mapzen extends AbstractHttpProvider
```
unkind: It makes sense to finalize this class.

vicchi: That's true; it probably does. However when I wrote this I used the existing provider implementations as examples of best practice and as far as I can see, none of these finalise their classes. Not that's not a good reason not to do this, but I would have thought it would make more sense to apply finalisation across the board, rather than have a single provider finalise and none of the others doing so.

Thoughts?

willdurand: > I would have thought it would make more sense to apply finalisation across the board, rather than have a single provider finalise and none of the others doing so

definitely yes!

vicchi: I've just opened a new issue (#537) so we track adding finalisation to all providers. But I'm also assuming that based on @willdurand's comment above, there's nothing stopping this PR from being merged now?

willdurand: I am +1

vicchi:
@willdurand - can this PR be merged now? I want to start working on the testing with encrypted provider API keys/credentials, but would prefer to have Mapzen tested as part of this.

willdurand:
thanks @vicchi!

vicchi:
Thanks!

stevevance:
Thank you for writing what I could not!
I'm going to test this, too. I prefer Mapzen Search because they aggregate results from OpenAddresses, OpenStreetMap, and Geonames.

stevevance:
I tested this finally and I got one unexpected result: the `adminLevel` number 2 returned the address's state in the USA. 

On the other providers that I use (ArcGISOnline, Bing, Nominatim) the `adminLevel` number 2 always returned the address's county.

src/Geocoder/Provider/Mapzen.php
unkind: So far this is the only provider without `final` keyword.

PR #552, opened 2016-11-19
Update unittest passage

arubacao:
Use phpunit from vendor to prevent issues
don't suggest to rename `phpunit.xml.dist` 
https://phpunit.de/manual/current/en/organizing-tests.html#organizing-tests.xml-configuration
If there is no `phpunit.xml` then phpunit will use `phpunit.xml.dist` , so whats the point in renaming it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -561,13 +561,12 @@ $ composer install --dev
 Then, run the following command:
 
 ```
-$ phpunit
+$ ./vendor/bin/phpunit
 ```
 
 You'll obtain some _skipped_ unit tests due to the need of API keys.
 
-Rename the `phpunit.xml.dist` file to `phpunit.xml`, then uncomment the
-following lines and add your own API keys:
+Uncomment the following lines in `phpunit.xml.dist` and add your own API keys:
 
 ``` xml
 <php>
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -561,13 +561,12 @@ $ composer install --dev
 Then, run the following command:
 
 ```
-$ phpunit
+$ ./vendor/bin/phpunit
 ```
 
 You'll obtain some _skipped_ unit tests due to the need of API keys.
 
-Rename the `phpunit.xml.dist` file to `phpunit.xml`, then uncomment the
```
Nyholm: One should rename the phpunit.xml.dist file to phpunit.xml to avoid accidentally checking in your API keys.

README.md
```diff
@@ -561,13 +561,12 @@ $ composer install --dev
 Then, run the following command:
 
 ```
-$ phpunit
+$ ./vendor/bin/phpunit
```
Nyholm: Even better: We should use a composer script so we can run:

```
composer test
```

README.md
arubacao: Makes sense, didn't see that in the gitignore. Elegant way to ensure security

README.md
arubacao: Will add this tomorrow

PR #555, opened 2016-11-24
Catch "Forbidden" errors with Bing maps

cweiske:
They can occur when a free but limited key is used and the number
no more requests are allowed.

### DIFF
--- a/src/Geocoder/Provider/BingMaps.php
+++ b/src/Geocoder/Provider/BingMaps.php
@@ -108,6 +108,10 @@ final class BingMaps extends AbstractHttpProvider implements LocaleAwareProvider
         }
 
         $json = json_decode($content);
+        if($json->statusCode == 403) {
+            $err = implode(", ", $json->errorDetails);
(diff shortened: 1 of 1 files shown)


willdurand:
Ideally, we should cover this change with a test case.

Nyholm:
I will close this PR. 

403 means forbidden, Authentication will not help. 
401 means that you are unauthenticated. 

Reference: https://msdn.microsoft.com/en-us/library/ff701703.aspx

Recent changes in master do already cover 401 erros.

cweiske:
@Nyholm - we saw the 403 status code in production. Our code silently failed because geocoder-php did not expose this error condition.

Catching 401 errors does not help in the circumstances described in the initial PR comment.

If you close this PR without handling 403 errors in any way, other people will fall in the same trap.

Nyholm:
Thank you.  
We are handling them. See https://github.com/geocoder-php/Geocoder/blob/master/src/Common/Provider/AbstractHttpProvider.php#L61

But we are not using there UnauthorizedException.

PR #563, opened 2016-12-22
Using PSR-4

Nyholm:
Stop using deprecated PSR-0 and remove al lot of directories not needed.


willdurand:
👍

Nyholm:
Thank you for the review

PR #567, opened 2016-12-23
feature: remove all non-https support

### LINKED ISSUE #562: Remove non-ssl support
The following adapters has options for using both SSL and non-secure transport. I think we should only support secure connections.

* [ ] MaxMind
* [ ] BingMaps
* [ ] Mapzen
* [ ] GoogleMaps
* [ ] OpenCage
* [ ] ArcGISOnline

Simperfit:
fix #562


src/Geocoder/Provider/ArcGISOnline.php
```diff
@@ -22,34 +22,29 @@
...
 
-    /**
-     * @var string
-     */
-    private $protocol;
 
     /**
      * @param HttpClient $client        An HTTP adapter
      * @param string     $sourceCountry Country biasing (optional)
      * @param bool       $useSsl        Whether to use an SSL connection (optional)
```
Nyholm: This comment should be removed

Simperfit:
@Nyholm test pass locally with php 5.6, do I have to do something about the coordinates ?

src/Geocoder/Provider/MaxMind.php
```diff
@@ -61,13 +57,12 @@
      * @param string     $service The specific Maxmind service to use (optional).
      * @param bool       $useSsl  Whether to use an SSL connection (optional).
```
Nyholm: This comment should be removed.

tests/Geocoder/Tests/Provider/GoogleMapsTest.php
```diff
@@ -196,60 +196,61 @@ public function testGeocodeBoundsWithRealAddressForNonRooftopLocation()
     public function testGeocodeWithRealAddressReturnsMultipleResults()
     {
         $provider = new GoogleMaps($this->getAdapter());
-        $results  = $provider->geocode('Paris');
+        $results  = $provider->geocode('i');
```
Nyholm: This line should not be changed. Keep "Paris" here and we do not have to modify the rest of the test.

Nyholm:
> test pass locally with php 5.6, do I have to do something about the coordinates ?

No, you should not change any of the coordinates or addresses. I think you accidentally modified the geocode string. I wrote an inline comment about that.

tests/Geocoder/Tests/Provider/GoogleMapsTest.php
Simperfit: If I let Paris, the test does not pass anymore because of https://maps.googleapis.com/maps/api/geocode/json?address=Paris%22 returning just one element

Nyholm: That is very strange... But the issue is at the geocoder service. Let's do as you do an remove 'Paris'.

Simperfit:
I think the ApiResult of Arcgis are inconsistent, since it does not break on the same things everytime

Nyholm:
True. Run your tests locally. Make sure that you have the following in your phpunit.xml
```
<server name="USE_CACHED_RESPONSES" value="true" />
```

If the test passes, make sure to checkin the new cached response that will be created in `tests/.cached_responses`

tests/Geocoder/Tests/Provider/GoogleMapsTest.php
Nyholm: Ah, you get new caches.. Please rename the file "tests/.cached_responses/0e41628de6d90aa50c426e12391617abc3ccf5ff" to "tests/.cached_responses/65760b8fc8d216f997a2e5b2ce499881426f7353".

Now it will work with "Paris"

Nyholm:
From https://github.com/geocoder-php/Geocoder/pull/567#discussion_r93773443
> Ah, you get new caches.. Please rename the file "tests/.cached_responses/0e41628de6d90aa50c426e12391617abc3ccf5ff" to "tests/.cached_responses/65760b8fc8d216f997a2e5b2ce499881426f7353".

> Now it will work with "Paris"

Simperfit:
@Nyholm done ;)

tests/Geocoder/Tests/Provider/GoogleMapsTest.php
```diff
@@ -213,26 +214,26 @@ public function testGeocodeWithRealAddressReturnsMultipleResults()
         /** @var Location $result */
         $result = $results->get(1);
         $this->assertInstanceOf('\Geocoder\Model\Address', $result);
-        $this->assertEquals(33.6609389, $result->getCoordinates()->getLatitude(), '', 0.001);
-        $this->assertEquals(-95.555513, $result->getCoordinates()->getLongitude(), '', 0.001);
+        $this->assertEquals(33.660938899999998, $result->getCoordinates()->getLatitude(), '', 0.001);
```
Nyholm: Why do you need to change these coordinates?

Simperfit: You are right, I don't need too ;)

Nyholm:
Thank you @Simperfit. Great work!

PR #569, opened 2016-12-23
Moved the rest of the providers to SSL

Nyholm:
I moved all providers to SSL (if the API supports it). I also make sure we test all providers. (ie, I uncomment the cache keys). Since we are using cache responses it will not matter if the key is valid or not.


willdurand:
👍

Nyholm:
Thank you for the review.

Issue #508, opened 2016-04-27
Misleading exception NoResult


pavolsenko:
Exception `NoResult` is a bit confusing. For example in `GoogleMaps` provider it's thrown when there is no/incorrect response from the server (no JSON returned) but also when there is a response but with zero results. It's message is also confusing saying "Could not execute query". This is not true. Query has been excuted correctly and returned valid JSON but this JSON is saying `ZERO_RESULTS`. This should be two different exceptions: one for invalid response from server and one for valid response but with zero results (`ZeroResults` exception? :wink: ). I've briefly checked other providers and they have similar problem with this exception. 
Thanks.

willdurand:
Yes, I do agree with you. Wanna contribute?

pavolsenko:
Sure :)

Nyholm:
Fixed by #572

FinesseRus:
@willdurand This issue is fixed for version 4.x which is still not stable and not released. So we still face the problem on a stable 3.x version. May you or I fix it for 3.x version?

Nyholm:
If this should be fixed for 3.x we need to consider BC. Or the invalid server response exception must extend zeroResultException. 
I currently focus all my efforts on finishing 4.0 before August, I do not have time to create such PR. But I would be happy to review a PR to the 3.x branch to fix this issue.

FinesseRus:
@Nyholm I understand, I will workaround it somehow on my side.

I can't understand this:

```php
class InvalidServerResponse extends ZeroResults
```

Why is `InvalidServerResponse` a subtype of `ZeroResults`? I believe that `InvalidServerResponse` is a sign of a problem that need to be fixed by a developer or administrator, and `ZeroResults` is a normal runtime situation (given address doesn't exist). I wouldn't extend one by another or I would just return an empty address collection in case of zero results. Am I wrong? Maybe I don't understand something.

Nyholm:
I agree with you. That is why it is changed in master: https://github.com/geocoder-php/Geocoder/blob/master/src/Common/Exception/InvalidServerResponse.php#L20

FinesseRus:
@Nyholm Oh, sorry, I didn't see it. Thank you for the answer!

PR #572, opened 2016-12-25
Make providers follow Liskov Substitution Principle

Nyholm:
We do not currently test that providers follow LSP. This PR will create such test and make sure to fix the issues shown by the test. 

WIP


Nyholm:
Questions: 
- When should NoResult exception be thrown? Some providers returns an empty collection and some throw the exception

- Should we force `Location::getStreetName` to return the street name without street number? Bing does not return the street name without the number. Google does...

I **suggest** that `Location::getStreetName` should never include number. I also suggest that we remove the NoResult exception. 

I need some feedback on these.

Nyholm:
Can I get some feedback from @geocoder-php/geocoder here?

Nyholm:
I've fixed #508 by introducing ZeroResults exception and InvalidServerResponse exception. I've also chosen to ignore the issue with street number.

Nyholm:
This looks like a major PR but it really isn't. It is just loads of test code and cached responses.

Issue #574, opened 2016-12-25
Our interfaces


Nyholm:
Currently we have the following interfaces for geocoder/providers: 

```php
interface LocaleAwareProvider extends Provider {
  public function getLocale();
  public function setLocale($locale);
}
interface Provider extends Geocoder {
  public function getName();
}
interface Geocoder {
  public function geocode($value);
  public function reverse($latitude, $longitude);
  public function getLimit();
  public function limit($limit);
}
```

Is there a reason why a Geocoder cannot have a name? If we should add support for a `Location::providedBy` (#490) I think all Geocoder should have a name. 

I would also like to hear voices about the interface inheritance. Must the `LocaleAwareProvider` really be a `Provider`?

I suggest the following: 

```php
interface LocaleAware {
  public function getLocale();
  public function setLocale($locale);
}

interface Geocoder {
  public function geocode($value);
  public function reverse($latitude, $longitude);
  public function getLimit();
  public function limit($limit);
  public function getName();
}
```

-------------

I suggest we start to add interfaces for features. In #501 I suggested to add a `BoundsAware`.
```php
interface BoundsAware {
  public function setBounds(Bounds $bounds);
}
```

I do also want an interface for providers that supports IP addresses. (Right now we have no way to tell)
```php
interface IpAddressGeocoder {
}
```

Nyholm:
Any thoughts? @toin0u @Baachi?

willdurand:
> Is there a reason why a Geocoder cannot have a name?

Yes. There is no need for it, but it requires more context to understand.

[...]

PR #576, opened 2017-01-03
Added BoundsAwareInterface

Nyholm:
As discussed in #574

### DIFF
--- a/src/Provider/BoundsAwareProvider.php
+++ b/src/Provider/BoundsAwareProvider.php
@@ -1,0 +1,24 @@
+<?php
+
+/**
+ * This file is part of the Geocoder package.
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @license    MIT License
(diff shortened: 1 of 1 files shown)


willdurand:
What happen if we have a provider that is both locale- + bound-aware?

Nyholm:
It has to implement two interfaces. You will run into an issue when you typehint for it:

```php
function (LocaleAwareProvider $provider) {
  if ($provider instanceof BoundsAwareProvider) {
    // Set bounds
  }
}
```

I do not know a better way of solving this. To create `Superinterface extends LocaleAwareProvider, BoundsAwareProvider` is wrong... To include all methods in one interface would violate interface segregation principle. 

Note that this is only an "issue" in third party libraries. An application developer would write: 

```php
function (ConcreateProvider $provider) {
    // Set bounds

}
```

Nyholm:
This is the wrong approach. We should use `GeocodeQuery::withBounds`

Issue #546, opened 2016-10-22
Google map Reverse Geocoding


ruscon:
https://developers.google.com/maps/documentation/geocoding/intro#ReverseGeocoding
These are the optional parameters: result_type and  location_type - can we add them ?

arubacao:
@Nyholm I'd also like this kind of functionality and I'm ready to send a PR for it.

Just some things to ask you upfront: 

1. `result_type` and `location_type` can be set through the constructor

```
public function __construct(HttpAdapterInterface $adapter, $locale = null, $region = null, $useSsl = false, $apiKey = null, $result_type = null, $location_type = null)
```
(as last parameters for backwards compatibility)

2.  `result_type` and `location_type` can be set through setters

3. Extend `buildQuery` method

3. Create new Exception class `InvalidRequestException` and throw it on `status` = `INVALID_REQUEST` responses

Please add your thoughts and comments & tell me if you approve this approach.
I'd appreciate a quick answer, so I can get started fast.

Cheers,
Chris

Nyholm:
This is now added. Have a look at #619 and #620

Nyholm:
Closed by #619 and #620

PR #583, opened 2017-02-19
Implement a Query object

Nyholm:
This is an implementation of a query object suggested in #574. Ping @unkind 

### Changes: 
* All providers are stateless (removed `LocaleTrait`)
* `Geocoder` extends `Provider`
* Removed `LocaleAwareProvider`
* Added `LocaleAwareGeocoder` (no methods)
* Added `StatefulGeocoder` (this will make migrating easy)
* Added `GeocodeQuery`, `ReverseQuery`
* Added `Provider::geocodeQuery`,`Provider::reverseQuery`

Only GoogleMaps and Bing are using the new Provider interface. 

### Why we need this
The Query objects keeps providers stateless. The query object is easily extendable with extra data like bounds, locale, limit and also arbitrary data. Allowing arbitrary data will solve #546. 

Using an query object over setters on the providers keeps the public interface of the providers more clean. 

This PR will also solve #501. 

### TODO: 
* [x] Make sure all providers are using the new Query objects
* [x] Update the tests
* [x] Update ProviderAggregator
* [x] Update change log


src/Model/Query/GeocodeQuery.php
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * @param Bounds $bounds
+     *
+     * @return GeocodeQuery
+     */
+    public function withBounds(Bounds $bounds)
+    {
+        $this->bounds = $bounds;
```
unkind: I suggest to make query object immutable, i.e. 
```php
$query = clone $this;
$query->bounds = $bounds;
```
`GeocodeQuery` is a value object.

src/Model/Query/Query.php
```diff
@@ -0,0 +1,26 @@
+<?php
+namespace Geocoder\Model\Query;
+
+use Geocoder\Model\Coordinates;
+
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+interface Query
```
unkind: I'm not sure if this interface is necessary.

src/Provider/BingMaps.php
```diff
@@ -94,25 +98,28 @@ public function getName()
...
 
     /**
-     * @param string $query
+     * @param Query $query
+     * @param string $url
+     *
+     * @return \Geocoder\Collection
      */
-    private function executeQuery($query)
+    private function executeQuery(Query $query, $url)
```
unkind: I'd pass 3 variables or even copy-paste URL generation for both queries.

src/Model/Query/Query.php
Nyholm: You are correct.

src/Model/Query/GeocodeQuery.php
Nyholm: I was thinking about making them immutable. I will make the query objects easier to reused and you will not get any unexpected behavior. Though, they are "builder". I compare the GeocodeQuery to Doctrines QueryBuilder. It would not make much sense to have the QueryBuilder immutable, would it? Or maybe Im missing something.

unkind: > It would not make much sense to have the QueryBuilder immutable, would it?

I'm not that familiar with Doctrine's QB, but as I can see:
* query builder is not perfect match for `GeocoderQuery`, query builder is not a query, it is a service; query, on other hand, is a message, queries should be compared by their values -> identity of query is based on value equality; 
* query builder probably was designed in the mutable way because of performance reasons, because cloning that object is more expensive than small object like `GeocoderQuery`; and number of method calls is much more higher than in your case: for complex query you may trigger `clone` 20-30 times very easy.

However, even for SQL query builder it probably makes sense to keep it immutable too.  Sometimes you need to execute very similar queries, but with little difference (e.g. simple query + the same with `COUNT(*)`). AFAIU, Doctrine suggests to `clone` QB explicitly when you need it, but it looks like a trade-off.

I'd rather compare your query with something like `Request` from PSR-7 if you are in doubt. Request and Query are very similar concepts. Whenever your modify your request/query/command/event, you create different message.

Nyholm: Okey, Thanks =)

Thanks for the review.

src/Model/Query/GeocodeQuery.php
```diff
@@ -0,0 +1,168 @@
...
+     * @return GeocodeQuery
+     */
+    public static function create($text)
+    {
+        return new self($text);
+    }
+
+    public function __clone()
+    {
+        $this->bounds = clone $this->bounds;
```
unkind: `Bounds` is an immutable structure: sharing same instance of `Bounds` can't be destructive.

Nyholm:
Thank you

Nyholm:
Today Ive done nothing to this PR but adding tests and making sure all providers implements the new interfaces.

src/GeocoderTrait.php
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function reverse($latitude, $longitude)
+    {
+        return $this->reverseQuery(ReverseQuery::fromCoordinates($latitude, $longitude));
+    }
+}
```
sagikazarmark: Missing newline

src/Model/Query/GeocodeQuery.php
```diff
@@ -0,0 +1,168 @@
+<?php
+
+/**
+ * This file is part of the Geocoder package.
```
sagikazarmark: Is this header mandatory? If so, it's missing from a few places.

src/Model/Query/GeocodeQuery.php
```diff
@@ -0,0 +1,168 @@
...
+     */
+    private $limit;
+
+    /**
+     * @var array
+     */
+    private $data;
+
+    /**
+     * @param Coordinates $coordinates
```
sagikazarmark: This does not seem to be correct

src/Model/Query/GeocodeQuery.php
```diff
@@ -0,0 +1,168 @@
...
+            throw new InvalidArgument('Geocode query cannot be empty');
+        }
+
+        $this->text = $text;
+        $this->data = [];
+        $this->limit = Provider::MAX_RESULTS;
+    }
+
+    /**
+     * @param $text
```
sagikazarmark: Type?

src/Model/Query/ReverseQuery.php
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * @param Coordinates $coordinates
+     *
+     * @return ReverseQuery
+     */
+    public static function create(Coordinates $coordinates)
+    {
+        return new self($coordinates);
+    }
+    /**
```
sagikazarmark: Missing newline

src/Model/Query/ReverseQuery.php
```diff
@@ -0,0 +1,142 @@
...
+     * @param float $latitude
+     * @param float $longitude
+     *
+     * @return ReverseQuery
+     */
+    public static function fromCoordinates($latitude, $longitude)
```
sagikazarmark: Personally I find this a bit more convenience stuff than necessary. Either create or fromCoordinates IMHO.

src/Model/Query/ReverseQuery.php
```diff
@@ -0,0 +1,142 @@
...
+    {
+        return new self(new Coordinates($latitude, $longitude));
+    }
+
+    /**
+     * @param int $limit
+     *
+     * @return $this
```
sagikazarmark: Should be the class name

src/Provider/FreeGeoIp.php
```diff
@@ -26,12 +28,13 @@
      * @var string
      */
     const ENDPOINT_URL = 'https://freegeoip.net/json/%s';
-
+
```
sagikazarmark: Indent

src/Provider/GoogleMaps.php
```diff
@@ -126,85 +121,97 @@ public function getName()
         return 'google_maps';
     }
 
+    /**
+     * @param $region
+     *
+     * @return GoogleMaps
+     */
     public function setRegion($region)
```
sagikazarmark: Is the set prefix preserved for BC reason?

src/Provider/TomTom.php
```diff
@@ -42,21 +43,20 @@
     /**
      * @param HttpClient $client An HTTP adapter.
      * @param string     $apiKey An API key.
-     * @param string     $locale A locale (optional).
      */
-    public function __construct(HttpClient $client, $apiKey, $locale = null)
+    public function __construct(HttpClient $client, $apiKey)
```
sagikazarmark: IIRC this was `$httpClient` a few thousand lines above.

src/StatefulGeocoder.php
```diff
@@ -0,0 +1,122 @@
...
+{
+    /**
+     * @var string
+     */
+    private $locale;
+
+    /**
+     * @var Bounds
+     */
+    private $bounds;
```
sagikazarmark: Currently this cannot be set at all

sagikazarmark:
I like the idea of immutability in Providers.

As for the queries: in case of the reverse query it seems to be a bit of overkill: basically coordinates are packed into an object, which gets unpacked in the provider as is. If it's a matter of consistency and/or probable future extension then it's probably fine.

Also, it feels weird that the limit can be passed in the query as well. One could say limit is a Provider configuration. I don't have too much context, so I might be wrong about it.

src/Provider/GoogleMaps.php
Nyholm: No, it is a typo

src/Provider/TomTom.php
Nyholm: I do not understand. We always refer to the `HttpClient` as `$client`

Nyholm:
Thank you for the review. Really appreciated. 

> As for the queries: in case of the reverse query it seems to be a bit of overkill: basically coordinates are packed into an object, which gets unpacked in the provider as is. If it's a matter of consistency and/or probable future extension then it's probably fine.

Yes, it is a matter of consistency and about being able to pass arbitrary data to the provider. 

> Also, it feels weird that the limit can be passed in the query as well. One could say limit is a Provider configuration. 

Hm, I've never considered that.. I compare it to Doctrine. The `Limit` it part of the MySQL query and not config.

src/StatefulGeocoder.php
Nyholm: Added a bunch of setter.

willdurand:
Good job!

PR #595, opened 2017-04-05
Update dependencies.

ddinchev:
While `egeloen/http-adapter` is deprecated and this dependency is removed in 4.0, ~0.8 version currently blocks me to use geocoder-php ~3.3 as another library (florianv/swap) requires `egeloen/http-adapter ~1.0`. At the same time `egeloen/http-adater ~1.0` does not seem to be backward incompatible with `0.8`. Currently `geocoder-php 4.0-dev` does not have frozen API so it feels risky to couple with it until at least we got a release candidate.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,10 +13,11 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "egeloen/http-adapter": "~0.8",
+        "egeloen/http-adapter": ">=0.8",
         "igorw/get-in": "~1.0"
     },
     "require-dev": {
+        "phpunit/phpunit": "^4.0",
         "geoip2/geoip2": "~2.0",
         "symfony/stopwatch": "~2.5"
     },


Nyholm:
Thank you for this PR. Two questions:

1) You say that "egeloen/http-adater ~1.0 does not seem to be backward incompatible with 0.8". Don't we need more chnages than just to update the composer.json? What where the updates between 0.8 and 1.0?

2) I know that the latest version of florianv/swap have dropped egeloen/http-adater and are using HTTPlug. Using the latest version of florianv/swap would also fix your issue, right?

egeloen:
For the record, bumping from 0.8 to 1.0 is pretty minimal as show by https://github.com/egeloen/ivory-http-adapter/compare/0.8.0...1.0.0 as I just have added support for Symfony 3, added a caching system and added a new adapter. Then, a move from 0.8 to 1.0 was just in order to release a stable release (no BC break)

ddinchev:
@Nyholm thanks for the swift reply! Probably I wasn't very clear - I meant it shouldn't cause any BC, so thanks @egeloen for clarifying! :) You are right - `florianv/swap ~3` did solve my issue. But I just updated my MR to actually require `>=0.8`, so to allow `1.x` as well.

Nyholm:
Sorry for forgetting about this PR.

Issue #578, opened 2017-01-03
Put all providers in separate packages


Nyholm:
Currently there are over 15 providers in this repo an at least 10 more in the geocoder-extra repo. I have a feeling that most developers do not use all providers available. I would like to suggest that we do a [subtree split](https://www.subtreesplit.com) of this repo to create the following repos

| Repository | Package name | Contents |
| -----------|----------------| --------- |
| geocoder-php/Geocoder | none | Everything |
| geocoder-php/php-common | willdurand/geocoder | Interfaces, Geocoders, Models, dumpers, exceptions, formatters |
| geocoder-php/google-maps-provider | geocoder-php/google-maps-provider | GoogleMaps provider + tests |
| geocoder-php/bing-maps-provider | geocoder-php/bing-maps-provider | BingMaps provider + tests |
| ... | ... | ... |
| geocoder-php/chain-provider | geocoder-php/chain-provider | Chain provider + tests |

To use one the GoogleMaps provider you would install `geocoder-php/google-maps-provider`.

### Benefits

* No optional dependencies or suggests
* No need for geocoder extra package
* Version each provider separately. If one provider has a BC break we do not need to bump all providers. 
* Statistics which provider is getting used

### Drawbacks

All providers must live in a separate namespace. That would be a BC break. 

### What we actually have to do

* We create a lots of new repos
* Move the test code to the providers
* Change the repository for package `willdurand/geocoder` to `geocoder-php/php-common`

### What do you think?

Is it a good idea? Should this be done before 4.0, 5.0 or not at all? Has 4.0 too many changes already?

-----------
Disclaimer: I am the author of https://www.subtreesplit.com so of course I think this is the best idea ever. Please give me some feedback.

arubacao:
I like the idea very much.
Another benefit:
- Additional third-party providers can be developed and maintained a lot easier (also from external people)

Just for inspiration purposes, Laravel has similar approaches:
https://socialiteproviders.github.io/
https://github.com/laravel-notification-channels

vingrad:
Yes, please make this. It's a good idea.

drupol:
Hi,

I'm for this kind of things of course.
The things is that it is a big refactoring.
I still believe that we should have tagged a release before doing this, just in case.

Nyholm:
We have currently some BC breaks in master so next tag will be 4.0.0. I hope that 4.0.0 will live a long life and 5.0 is in late 2018. If we wait with this feature we will not see it until 2018.

[...]

PR #604, opened 2017-04-28
Move providers to new namespace

Nyholm:
This is a first PR for #578

I've just done the Chain provider the moment. All other providers will follow the same pattern. 

#### TODO
 * [x] Move the rest of the providers
 * [x] Update change log
 * [x] Make sure travis runs the tests


Nyholm:
Im done with this PR. Deps low will fail because it is using code in master.

PR #610, opened 2017-05-13
Moved Queries away from the model namespace

Nyholm:
Moved queries to Geocoder\Query instead of Geocoder\Model\Query. 

The models should be something that could change depending on provider, the queries does not


Nyholm:
Ping @svrdlic. What do you think about moving the queries from `Geocoder\Model\Query` to `Geocoder\Query`?

svrdlic:
I agree with you @Nyholm and I think that folder structure will be better organized with this changes

Nyholm:
Thank you for the review.

PR #613, opened 2017-05-13
Review exceptions

Nyholm:
I've removed the following exceptions `HttpError`, `UnexpectedValue`  because they were not used. 
I've also removed  `CollectionIsEmpty` because an AddressCollection::first should never throw an exception. See Doctrine's ArrayCollection. 

I've also removed `ZeroResults`, `ChainZeroResults`. Throwing an exception when we do not get a results is strange. Compare it to a SQL statement. A zero result is also a result, it is not an exception. However, if the server fails somehow, then we should throw an exception ie `InvalidServerResponse`.


unkind:
> I've also removed CollectionIsEmpty because an AddressCollection::first should never throw an exception. See Doctrine's ArrayCollection.

I'm not sure I understand this argumentation. Because Doctrine is always correct?

Nyholm:
Thank you @unkind for taking your time to review this PR. No, Doctrine is not **always** correct. However there are some very smart people working with Doctrine, the project is far more used then this Geocoder library. It does also make sense to adapt to others. That will improve our DX and reduce the number of WTF for our users. 

It will also comply with the behavior of PHP's `reset` and `end`.

What is your opinion? Do you want to throw an exception? If so, what is your reasoning?

willdurand:
Mmh I barely remember that the ZeroResults exceptions covered a concrete use case but I di not remember it... Poke @toin0u @Baachi

unkind:
> If so, what is your reasoning?

The same reasoning why `get()` throws `OutOfBoundsException`.
If you need to check emptiness, you can call `count()` or add `empty(): bool`.
`null` makes result ambiguous and introduces unnecessary checks.
Sometimes I don't care about cases when something doesn't exist. I'd appreciate if library would throw exception, so I can log it. But if library returns `null` it makes things harder: first, it fails with `Fatal error` with PHP < 7; second, it may point to the "wrong" line:
```
$address = $addressCollection->first(); // Instead of exception here ...
$address->...; // I'd get \EngineException here
```

Nyholm:
Good. I like your suggestion @unkind. I will add an `empty():bool` to the collection's interface and make sure `first` return a `Location` or throws an exception.

src/Common/Exception/CollectionIsEmpty.php
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @license    MIT License
+ */
+
+namespace Geocoder\Exception;
+
+/**
+ * When you are trying to access an element on en empty collection.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
```
unkind: Offtopic: https://www.databasesandlife.com/dont-use-the-author-javadoc-tag/ (I've noticed the original class had no `@author` tag: it proves the point that the tag is useless).

src/Common/Exception/CollectionIsEmpty.php
```diff
@@ -10,6 +10,11 @@
 
 namespace Geocoder\Exception;
 
-class CollectionIsEmpty extends \RuntimeException implements Exception
+/**
+ * When you are trying to access an element on en empty collection.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
```
unkind: Offtopic: https://www.databasesandlife.com/dont-use-the-author-javadoc-tag/

Nyholm: Thank you for the review. I know it does not have lots of importance. I like to use this as "Who knows the most of this class". It will make it easier to ping someone if one has questions. Though, yes it is silly in this case. I included it because it is a standard class comment of mine.

Nyholm:
Thanks again for the review. It has not been any major changes on this PR, I will merge this in order to move forward with our 4.0 release. 

If @toin0u, @Baachi or anyone else see any issues later, feel free to open an issue or a new PR.

PR #617, opened 2017-05-15
Create a common function to fetch data form server

Nyholm:
This will be a unified way to throw some exceptions

This will reduce the number of checks in each provider.


Nyholm:
Im done with this PR now. I will not catch HttpException from the HttpClient. If one adds the ErrorPlugin they probably want to have HttpException thrown.

willdurand:
👍

PR #618, opened 2017-05-18
Interface change docs fix.

ddinchev:
This PR is more to assert that this is the intended interface change in v4 than to contribute a very meaningful change.

### DIFF
--- a/src/Common/Geocoder.php
+++ b/src/Common/Geocoder.php
@@ -33,7 +33,7 @@ interface Geocoder extends Provider
      *
      * @param string $value
      *
-     * @return Collection
+     * @return \Geocoder\Model\Query\GeocodeQuery
      *
      * @throws \Geocoder\Exception\Exception
      */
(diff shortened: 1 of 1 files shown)


ddinchev:
Actually I think it's not the right time. @Nyholm I now see that you are in the middle changing namespaces. 

My confusion comes from the fact that the `Geocoder` interface is implemented by both the `ProviderAggregator` and the specific providers. While the `ProviderAggregator` would still return a `Collection`, the specific providers would return a `Query`. Is it planned to change?

Nyholm:
Hm, The Geocoder interface should not be implemented by any `Provider`. That is wrong. Which one is implementing `Geocoder` interface?

The `ProviderAggregator` is a geocoder, it should return a `Collection` whenever you call ´geocode´, `reverse`, `geocodeQuery` or `reverseQuery`. No provider will **return** a `Query`. The `Query` is used as an input to the `Providers`.

PR #543, opened 2016-10-20
Add level to Address::toArray function

AntoineLemaire:
Fix #516

PR #622, opened 2017-05-20
Add level to Address::toArray function

Nyholm:
This will replace #543


Nyholm:
I will merge this now since it is just a rebase of #543. That PR had plenty of reviews already.

Issue #513, opened 2016-07-09
Geocode precision type for Google and Yandex


acidmk:
Is it possible to get Yandex "GeocoderMetaData.precision" and Google "geometry.location_type" attributes from address results ?

willdurand:
I am not sure because we would like to have the same result sets over all the providers.

Nyholm:
Related to #533

Nyholm:
Closed by #624

PR #624, opened 2017-05-20
Extend result from Yandex

### LINKED ISSUE #532: Impossible to get subway station name in Yandex provider
When you requesting reverse geocoding with Yandex provider, using a 'metro' toponym, you can't get station name.  Original API URL: https://geocode-maps.yandex.ru/1.x/?geocode=30.324285,%2060.036843&kind=metro&format=json
You can't get "PremiseName" using Geocoder now.

Nyholm:
This will fix #532 and #513

### DIFF
--- a/src/Provider/Yandex/Model/YandexAddress.php
+++ b/src/Provider/Yandex/Model/YandexAddress.php
@@ -23,6 +23,13 @@ class YandexAddress extends Address
     private $precision;
 
     /**
+     * The name of this location.
+     *
+     * @var string|null
+     */
+    private $name;
+
+    /**
      * @return null|string
      */
     public function getPrecision()
@@ -38,6 +45,26 @@ class YandexAddress extends Address
     public function setPrecision($precision)
     {
         $this->precision = $precision;
+
+        return $this;
+    }
+
+    /**
+     * @return null|string
+     */
+    public function getName()
+    {
+        return $this->name;
+    }
+
+    /**
+     * @param null|string $name
+     *
+     * @return YandexAddress
+     */
+    public function setName($name)
+    {
+        $this->name = $name;
 
         return $this;
     }
--- a/src/Provider/Yandex/Model/YandexAddress.php
+++ b/src/Provider/Yandex/Model/YandexAddress.php
@@ -1,0 +1,44 @@
+<?php
+
+/*
+ * This file is part of the Geocoder package.
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @license    MIT License
+ */
+
+namespace Geocoder\Provider\Yandex\Model;
+
+use Geocoder\Model\Address;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class YandexAddress extends Address
+{
+    /**
+     * @var string|null
+     */
+    private $precision;
+
+    /**
+     * @return null|string
+     */
+    public function getPrecision()
+    {
+        return $this->precision;
+    }
+
+    /**
+     * @param null|string $precision
+     *
+     * @return YandexAddress
+     */
+    public function setPrecision($precision)
+    {
+        $this->precision = $precision;
+
+        return $this;
+    }
+}
--- a/src/Provider/Yandex/Yandex.php
+++ b/src/Provider/Yandex/Yandex.php
@@ -165,6 +165,7 @@ final class Yandex extends AbstractHttpProvider implements LocaleAwareGeocoder, 
             /** @var YandexAddress $location */
             $location = $builder->build(YandexAddress::class);
             $location->setPrecision(isset($flatArray['precision']) ? $flatArray['precision'] : null);
+            $location->setName(isset($flatArray['name']) ? $flatArray['name'] : null);
             $locations[] = $location;
 
         }
--- a/src/Provider/Yandex/Yandex.php
+++ b/src/Provider/Yandex/Yandex.php
@@ -167,7 +167,6 @@ final class Yandex extends AbstractHttpProvider implements LocaleAwareGeocoder, 
             $location->setPrecision(isset($flatArray['precision']) ? $flatArray['precision'] : null);
             $location->setName(isset($flatArray['name']) ? $flatArray['name'] : null);
             $locations[] = $location;
-
         }
 
         return new AddressCollection($locations);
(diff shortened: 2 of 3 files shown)


src/Provider/Yandex/Yandex.php
```diff
@@ -116,57 +118,57 @@ private function executeQuery($query, $locale, $limit)
...
-                'bounds' => $bounds,
-                'streetNumber' => isset($details['PremiseNumber']) ? $details['PremiseNumber'] : null,
-                'streetName' => isset($details['ThoroughfareName']) ? $details['ThoroughfareName'] : null,
-                'subLocality' => isset($details['DependentLocalityName']) ? $details['DependentLocalityName'] : null,
-                'locality' => isset($details['LocalityName']) ? $details['LocalityName'] : null,
-                'adminLevels' => $adminLevels,
-                'country' => isset($details['CountryName']) ? $details['CountryName'] : null,
-                'countryCode' => isset($details['CountryNameCode']) ? $details['CountryNameCode'] : null,
-            ]);
+            $builder->setStreetNumber(isset($flatArray['PremiseNumber']) ? $flatArray['PremiseNumber'] : null);
```
willdurand: if we go for PHP > 7, we might want to use `??`

Nyholm: Yeah, I agree. But should PHP >= 7 be a requirement for 4.0?

Nyholm: See https://github.com/geocoder-php/Geocoder/issues/634

willdurand: I think so. It is time :p

Nyholm: Wohoo!

src/Provider/Yandex/Model/YandexAddress.php
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+namespace Geocoder\Provider\Yandex\Model;
+
+use Geocoder\Model\Address;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class YandexAddress extends Address
```
unkind: `final`?

src/Provider/Yandex/Model/YandexAddress.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        return $this->precision;
+    }
+
+    /**
+     * @param null|string $precision
+     *
+     * @return YandexAddress
+     */
+    public function setPrecision($precision)
```
unkind: VO with setter?

Nyholm: Yes. I agree with you. But I'm not sure how to make it work without a setter. Look at the LocationBuilder.  Do you have any suggestions?

src/Provider/Yandex/Model/YandexAddress.php
Nyholm: Yes. I'll make sure to send a PR. Thanks

src/Provider/Yandex/Model/YandexAddress.php
unkind: I don't see much sense in `LocationBuilder`, I'd write something like

```php
$adminLevels = ...;

$locations[] = new YandexLocation(
    isset($flatArray['name']) ? $flatArray['name'] : null,
    isset($flatArray['precision']) ? $flatArray['precision'] : null,
    $adminLevels,
    ...
);
```

Nyholm: The LocationBuilder is meant to remove that large array all providers were building. 

Your suggestion means that the (already big and ugly) constructor of Address has to be overridden. Is that somehow avoidable? Could we make the constructor prettier?

unkind: > Your suggestion means that the (already big and ugly) constructor of Address has to be overridden. Is that somehow avoidable? Could we make the constructor prettier?

OK, it is "big and ugly". What's the problem with it? If `Address` requires a lot of params, it is natural that constructor is "big and ugly". Setter doesn't make code "prettier", it just makes object less explicit, it hides the problem (if it does exist).

If you ask me, many arguments for data object is OK, it is not a service.

Nyholm: Okey, Thank you. But how to call the constructor to subclasses of `Address` with the `LocationBuilder`? (which should be renamed to `AddressBuilder`)

unkind: [Wikipedia](https://en.wikipedia.org/wiki/Builder_pattern) states that the builder pattern solves the the [telescoping constructor anti-pattern](http://www.captaindebug.com/2011/05/telescoping-constructor-antipattern.html#.WSgLeFKB1-0) problem. C#, for example, in the past had no default values for properties and default values for arguments, so there were 2 ways to make default values: constructor overloading and the builder pattern. Regarding to your case with `Address` I don't see why you need the builder.

Technically, it just replaces
```php
$a = ...;
$b = ...;
$c = ... ? ... : null;

$address = new Address($a, $b, $c);
```
with
```php
$builder->setA(...);
$builder->setB(...);
$builder->setC(... ? ... : null);

$address = $builder->build();
```

What problem does the `LocationBuilder` solve?

Nyholm: The problem is that all providers is not setting all data. Some data might me permanently ignored by some providers. (`MaxMindBinary` is a perfect example). The `LocationBuilder` brings an easy way (with setter/typehints) to add some/all data. The logic (order of constructor parameters) is hidden away. Before the `LocationBuilder` was introduces we used an assoc array. 

Example:
```php
$builder->setE($httpResponse['e']);
$address = $builder->build();
```
Compared with: 
```php
$e = $httpResponse['e'];
$address = new Address(null, null, null, null, $e);
```

-------------

> But how to call the constructor to subclasses of Address with the LocationBuilder?

unkind:
> Compared with:

Why not? Provider makes translation from external API to your generic `Address` class, it is OK to have some dirty code here, isn't it? Offtopic: https://wiki.php.net/rfc/named_params

If you can't avoid some setter-like method, I suggest to use `with`-er syntax:

```php
$address = $address->withFoo('bar'); // with* clones the object and return changed one
```
At least, it makes object immutable.

src/Provider/Yandex/Model/YandexAddress.php
unkind: > Compared with:

Why not? Provider makes translation from external API to your generic `Address` class, it is OK to have some dirty code here, isn't it? Offtopic: https://wiki.php.net/rfc/named_params

If you can't avoid some setter-like method, I suggest to use `with`-er syntax:

```php
$address = $address->withFoo('bar'); // with* clones the object and return changed one
```
At least, it makes object immutable.

PR #628, opened 2017-05-20
Allow to set zoom

### LINKED ISSUE #598: Nominatim provider
Why is it not possible to set the zoom within the constructor of the nominatim (OpenStreetmap) provider?

Nyholm:
This will fix #598

This PR will also update `*Query::getData` to be more DX.

### DIFF
--- a/src/Common/Query/GeocodeQuery.php
+++ b/src/Common/Query/GeocodeQuery.php
@@ -156,11 +156,24 @@ final class GeocodeQuery
     }
 
     /**
-     * Return arbitrary data for this query.
+     * @param string $name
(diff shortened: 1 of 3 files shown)


Nyholm:
Ping @svrdlic. What do you think of the `getData($name, $default = null)` and `getAllData()` functions? Do they make sense? Is there a better name than `getAllData()` ?

willdurand:
:+1:

svrdlic:
Sorry for late response. Function names are very self-descriptive.. getAllData() to get array of data, and getData($name, $default=null) to get concrete data value, sounds great to me.

PR #633, opened 2017-05-20
Added factory method to Address.

Nyholm:
This will completely remove the need from LocationFactory.
The factory method will be used in our tests.

Im not sure we do need a factory class at all. Please review.


willdurand:
looks good to me!

Nyholm:
Thank you for the review

PR #635, opened 2017-05-22
Make sure we use PHP7 and PHPunit 6

Nyholm:


Nyholm:
Symfony will drop HHVM support, So can we. 
It was some errors with composer. When specifying PHP >=7 then it could not be installed on HHVM...

PR #655, opened 2017-05-29
adding support for formatted_address for GoogleMaps

svrdlic:
Just added support for 'formatted_address' for GoogleMaps.


svrdlic:
Thanks! :champagne:

Issue #593, opened 2017-03-20
Can `Address Line 2` field (or similar) be added?


mikebronner:
@ellisio posted in the GeocoderLaravel issues list (https://github.com/geocoder-php/GeocoderLaravel/issues/52) the following request:

> When parsing an address with an apartment number, you receive a "Subpremise" attribute. This attribute is typically mapped to an "Address Line 2".
>
> For example: https://maps.googleapis.com/maps/api/geocode/json?address=2123%20W%20Mineral%20Ave%20Apt%2061,%20Littleton,%20CO%2080120
>
> It would be nice to have this added as the subpremise on the Address object.

Is this something you would consider adding? Or any thoughts on work-arounds?

Thanks!
~Mike

Nyholm:
Does it make sense to add "getSubpremise" to the GoogleAddress? When/how often does subpremise exist in the response from google?

Ping @svrdlic

svrdlic:
From google maps docs: subpremise indicates a first-order entity below a named location, usually a singular building within a collection of buildings with a common name.

‘Subpremise’ is what Google Maps return in case that query address contains Apt/Suite/Unit number. I think it wont hurt if we add this, if someone adds it in query, it would be nice to get it back :)

Nyholm:
Okey, good. 

Can you submit a PR that adds supremise to the `GoogleAddress`?

svrdlic:
Sure

Nyholm:
Closed by #657

PR #657, opened 2017-06-02
Adding support for subpremise for GoogleMaps

svrdlic:
Added support for 'subpremise' for GoogleMaps, related to issue #593

### DIFF
--- a/src/Provider/GoogleMaps/GoogleMaps.php
+++ b/src/Provider/GoogleMaps/GoogleMaps.php
@@ -257,8 +257,8 @@ final class GoogleMaps extends AbstractHttpProvider implements LocaleAwareGeocod
             if (isset($result->formatted_address)) {
                 $address = $address->withFormattedAddress($result->formatted_address);
             }
-            if (property_exists($builder, 'subpremise')) {
-                $address = $address->withSubpremise($builder->subpremise);
+            if ($builder->hasValue('subpremise')) {
+                $address = $address->withSubpremise($builder->getValue('subpremise'));
             }
             $results[] = $address;
 
@@ -315,7 +315,7 @@ final class GoogleMaps extends AbstractHttpProvider implements LocaleAwareGeocod
                 break;
 
             case 'subpremise':
-                $builder->subpremise = $values->long_name;
+                $builder->setValue('subpremise',$values->long_name);
                 break;
 
             default:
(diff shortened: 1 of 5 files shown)


src/Provider/GoogleMaps/GoogleMaps.php
```diff
@@ -257,6 +257,9 @@ private function fetchUrl($url, $locale, $limit, $region)
             if (isset($result->formatted_address)) {
                 $address = $address->withFormattedAddress($result->formatted_address);
             }
+            if (property_exists($builder, 'subpremise')) {
```
Nyholm: It seams like you are checking the $builder if the "subpremise" exists. Shouldn't we check the result from google?

src/Provider/GoogleMaps/GoogleMaps.php
```diff
@@ -311,6 +314,10 @@ private function updateAddressComponent(AddressBuilder $builder, $type, $values)
                 $builder->setSubLocality($values->long_name);
                 break;
 
+            case 'subpremise':
+                $builder->subpremise = $values->long_name;
```
Nyholm: Not sure we can do this. `$builder->subpremise` is not public. One should aways use setters.

src/Provider/GoogleMaps/GoogleMaps.php
svrdlic: I dynamically added 'subpremise' property to AddressBuilder class, because that class is shared among other providers and contains properties that are common for all providers. 
That's why I didn't add it as filed to the class, and it doesn't have setter..  
I'm checking if 'subpremise' exists in $builder, because if it exists in result, then it is dynamically added to $builder object.

Nyholm: I see, this is quite nice. But property overloading like this is quite slow. 

Lets add a new property to the `AddressBuilder` that can take arbitrary data. We also need `AddressBuilder::setValue($name, $value)`, `AddressBuilder::hasValue($name)`, `AddressBuilder::getValue($name, $default=null)`.

svrdlic: Thank you. This is great, I will update code.. and add a test.

Nyholm:
Great, It all looks good. Only one thing. When you run the test, a new file in src/Provider/GoogleMaps/Tests/.cached_results will be created. Please check in that file too.

svrdlic:
Is it ok now? 
Could you tell me why we have untracked cache responses in tests/.cache_response?

Nyholm:
You made a new test with a new request that is why you had a new cached request.

svrdlic:
I understand that part, but I have new files for Yandex.ru, FreeGeoIp, with regular responses in /tests/.cache_response... Should we add them? The rest of untracked files are error messages about wrong api key.

PR #660, opened 2017-06-03
Use PHP7 syntax

Nyholm:


Nyholm:
This is blocking #658, #659, #661.

PR #479, opened 2015-12-23
Added locale support

shtricul:
Tested on 'ru', 'de' locale

PR #676, opened 2017-06-10
Added locale aware support for FreeGeoIp provider

shtricul:
Make support of locale aware for FreeGeoIp provider

The beginning of discussion is [here](https://github.com/geocoder-php/Geocoder/pull/479).

```php
$provider = new FreeGeoIp($adapter);

$provider->setLocale('fr');
$results = $provider->geocodeQuery(GeocodeQuery::create('81.27.51.251'));
/* or */
$results = $provider->geocodeQuery(GeocodeQuery::create('81.27.51.251')->withLocale('fr'));
```

### DIFF
--- a/src/Http/Provider/AbstractHttpProvider.php
+++ b/src/Http/Provider/AbstractHttpProvider.php
@@ -59,6 +59,7 @@ abstract class AbstractHttpProvider extends AbstractProvider
     protected function getUrlContents(string $url): string
     {
         $request = $this->getRequest($url);
+        $response = $this->getHttpClient()->sendRequest($request);
 
         $statusCode = $response->getStatusCode();
         if (401 === $statusCode || 403 === $statusCode) {


src/Common/Provider/AbstractHttpProvider.php
```diff
@@ -53,9 +54,9 @@ public function __construct(HttpClient $client, MessageFactory $factory = null)
      *
      * @throws InvalidServerResponse
      */
-    protected function getUrlContents($url)
+    protected function getUrlContents(string $url): string
```
Nyholm: Good!

src/Provider/FreeGeoIp/FreeGeoIp.php
```diff
@@ -83,4 +94,36 @@ public function getName(): string
     {
         return 'free_geo_ip';
     }
+
+    /**
+     * @param string $locale
+     */
+    public function setLocale(string $locale)
```
Nyholm: This makes the provider stateful. That is something we should avoid in 4.0.  I suggest adding this as constructor parameter. 

Also if the Query object has a locale parameter we should override the `$this->locale`

shtricul: I have also a proposal. I think that a second parameter in Common\Provider\AbstractHttpProvider is redundant and he is always overwritten in child classes. And I think this violates the Liskov Substitution Principle.

Nyholm: I can understand that you feel it is pointless to have the `MessageFactory` as a second parameter. But that is a problem with the child classes but not the `AbstractHttpProvider`.

Feel free to open a PR about it. 

> And I think this violates the Liskov Substitution Principle.

No it does not, since the constructor is not a part of the `Provider` contract

src/Common/Provider/AbstractHttpProvider.php
```diff
@@ -35,13 +36,11 @@
     private $messageFactory;
 
     /**
-     * @param HttpClient          $client
-     * @param MessageFactory|null $factory
+     * @param HttpClient $client
      */
-    public function __construct(HttpClient $client, MessageFactory $factory = null)
+    public function __construct(HttpClient $client)
```
Nyholm: Please make this change in a separate issue/PR. It requires far more discussion than your "add locale feature"

src/Provider/FreeGeoIp/FreeGeoIp.php
```diff
@@ -50,6 +70,10 @@ public function geocodeQuery(GeocodeQuery $query): Collection
             ])]);
         }
 
+        if ($locale = $query->getLocale()) {
```
Nyholm: We should not change the providers locale each time we do a geocode lookup. That will make the provider stateful.

src/Provider/FreeGeoIp/FreeGeoIp.php
```diff
@@ -83,4 +107,28 @@ public function getName(): string
     {
         return 'free_geo_ip';
     }
+
+    /**
+     * @return string|null
+     */
+    public function getLocale()
```
Nyholm: Why is this needed?

Nyholm:
Ping

Nyholm:
Ping?

Nyholm:
If you want this to be included in the 4.0.0 release you must update this PR so I can review it and merge it before the end of the month.

Nyholm:
Make sure you do a rebase on this PR as well to avoid the conflicting files.

Nyholm:
It looks like you merged things instead of rebasing. To recover from this I think it will be easiest to checkout and pull master and then cherry pick the commits you like. 

```bash
git checkout -b feegeoip-locale origin/master
git cherry-pick 96a2541
git cherry-pick d8e8de6
. . .
git cherry-pick dcc7800
```

If you get conflicts when cherry picking you need to solve them. 

The other approach is to check out a new branch and then copy your changes. 

```bash
git checkout -b feegeoip-locale origin/master
```

shtricul:
Sorry, have a lot of conflicts. So I re-fork and cherry-pick from old commits and open a new PR.

PR #693, opened 2017-06-17
Removed dependency in igorw

### LINKED ISSUE #686: Remove dependency on igorw/get-in
We barely use is in Common

Nyholm:
This will fix #686


Nyholm:
Thank you for the review.

PR #705, opened 2017-06-17
All exceptions should be ours

Nyholm:
The `\LogicException` in `AddressBuilder` will only be thrown if the provider is doing something strange (ie wrong programming).

How about the `\OutOfBoundsException` in `AddressCollection` and `AdminLevelCollection`. Should we create our own `OutOfBoundsException`?

### DIFF
--- a/src/Common/Assert.php
+++ b/src/Common/Assert.php
@@ -12,6 +12,8 @@ declare(strict_types=1);
 
 namespace Geocoder;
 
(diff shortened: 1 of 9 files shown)


willdurand:
We should have our own exceptions for  everything, ideally with a common interface to quickly catch all Geocoder exceptions.

Nyholm:
Thank you for the review

PR #707, opened 2017-06-17
Moved code of conduct to a separate file

Nyholm:
... to comply with github standards.


willdurand:
Yep, I saw that with their new tool last week. :+1:

Nyholm:
I just found it =)

Can you please give your thoughts about #705. Apart form that I think we are done =) I will just let the betas sit for a 5-6 weeks.

Issue #727, opened 2017-07-01
Making the `ProviderAggregator` final makes it harder to test


ddinchev:
In my application, I have a thin wrapper around geocoder, that would depend on `Geocoder\ProviderAggregator`. I'd test that by mocking it. However, mocking final classes is not possible (not with the defacto standard `phpunit-mock-objects` at least). So after updating to the latest master, my tests now break.

What is the benefit or goal of having the `ProviderAggregator` to be final? Could we also extract an Aggregator interface?

Nyholm:
Thank you for this issue. 

I usually make all classes final as it is much easier to remove a `final` than adding one. For this specific class there are no real reasons. You can send a PR to remove it if you want.

May I ask why you need to extend the `ProviderAggregator`?

ddinchev:
I don't - just not having an interface makes it a requirement to mock the class itself. And final classes are not mockable.

One potential case for extending (or providing an alternative implementation of such aggregator) is such where you would have a certain logic to automatically select a provider based on your query. For example -> use `geoip2` for valid IP address `geocode()` calls, use `google maps` for anything else. Given the popularity of the library, I think more options to extend the library are better. I'd be happy with an interface covering the public API of the class as well 👍

Nyholm:
That is an cool idea. Let me make a PR for that.

Nyholm:
Thank you @ddinchev. 

I will close this PR. `ProviderAggregator` is not final any more and your specific use case is now supported in the library.

PR #729, opened 2017-07-01
Mark ProviderAggregator as NOT final.

ddinchev:
Marking classes as `final` does have benefits in certain situations but many sources recommend doing so only when:
* There is an abstraction (interface) that the final class implements
* All of the public API of the final class is part of that interface

The particular thing that broke in my case was that it's not possible to mock `ProviderAggregator` anymore (and I'm testing the thin wrapper I wrote around it) as mocking final classes in `phpunit-mock-objects` (the default mocking library for phpunit) is not possible.

I would be happy (and could open a PR with such change) if we extract an inteface from the `ProviderAggregator` and leave it final.

Addresses #727

### DIFF
--- a/src/Common/ProviderAggregator.php
+++ b/src/Common/ProviderAggregator.php
@@ -21,7 +21,7 @@ use Geocoder\Provider\Provider;
 /**
  * @author William Durand <william.durand1@gmail.com>
  */
-final class ProviderAggregator implements Geocoder
+class ProviderAggregator implements Geocoder
 {
     /**
      * @var Provider[]


unkind:
> The particular thing that broke in my case was that it's not possible to mock ProviderAggregator anymore

This class is so simple so it could not be mocked at all.
At least, I'd suggest to mark such classes with `@final` phpdoc as they are final by themselves, mocking is just infrastructure issue for me.

unkind:
If you need some _argument from authority_, check out post of Mathias: http://verraes.net/2014/05/final-classes-in-php/
> Making a class final has some drawbacks. For example, you cannot mock it anymore with PHPUnit. This is a limitation of the infrastructure. One option is to remove the final keyword, but add an `@final` annotation.

PR #730, opened 2017-07-02
Geonames: Add countryInfo endpoint

SrihariThalla:
Endpoints:
* [x] countryInfo

Finalize:
* [x] Unit tests
* [x] Rebase
* [x] StyleCI fixes


src/Provider/Geonames.php
```diff
@@ -94,6 +104,89 @@ public function getName()
...
+
+        return $this->executeQuery($query);
+    }
+
+    /**
+     * Obtain the most population cities of a Country
+     *
+     * @param countryCode  It's an identifier on a Country Object in Geonames
+     */
+    public function getAllCitiesByCountryCode($countryCode)
```
SrihariThalla: Would this be helpful?

src/Provider/Geonames/Geonames.php
```diff
@@ -92,6 +97,43 @@ public function reverseQuery(ReverseQuery $query): Collection
...
+     *
+     * @param int|null $maxRows
+     *
+     * @param int|null $startRow
+     *
+     * @return Collection
+     *
+     * @throws \Geocoder\Exception\Exception
+     */
+    public function getCountryInfo(string $country = null, string $locale = null, int $maxRows = null, int $startRow = null): Collection
```
SrihariThalla: Shall I move these parameters to a Model class?

src/Provider/Geonames.php
```diff
@@ -0,0 +1,269 @@
+<?php
```
Nyholm: Why did you put this file here? 
Consider moving it to src/Providers/Geonames/

It may be an issue with you not using the latest master.

SrihariThalla: Yes, it's because of not using the latest master. I've moved the functionality to correct file.

I'd like to add just `getCountryInfo` in this PR. I'll add more methods in another PR.

Nyholm:
Could you address the style fixes suggested by StyleCI?

SrihariThalla:
Sure @Nyholm I'll update the style fixes and push tests.

src/Provider/Geonames/Geonames.php
```diff
@@ -164,4 +206,56 @@ private function executeQuery(string $url, string $locale = null): AddressCollec
...
         return new AddressCollection($results);
     }
+
+    /**
+     * @param string      $url
+     * @param string|null $locale
+     *
+     * @return AddressCollection
+     */
+    private function executeExtendedQuery(string $url, string $locale = null)
```
Nyholm: Is this function needed? Can we just make `getCountryInfo` longer?

src/Provider/Geonames/Geonames.php
```diff
@@ -92,6 +97,43 @@ public function reverseQuery(ReverseQuery $query): Collection
...
+     * @return Collection
+     *
+     * @throws \Geocoder\Exception\Exception
+     */
+    public function getCountryInfo(string $country = null, string $locale = null, int $maxRows = null, int $startRow = null): Collection
+    {
+        $url = sprintf(self::BASE_ENDPOINT_URL, 'countryInfoJSON', $this->username);
+
+        if (isset($country))
+        {
```
Nyholm: This is not PSR2. See StyleCI

src/Provider/Geonames/Geonames.php
```diff
@@ -92,6 +97,43 @@ public function reverseQuery(ReverseQuery $query): Collection
     }
 
     /**
+     * @param string|null $country
+     *
+     * @param string|null $locale
+     *
```
Nyholm: No space here

src/Provider/Geonames/Geonames.php
Nyholm: Should we really allow $country to be null?

SrihariThalla: It is not passed, Geonames returns all countries [subjected to the limit (Default: 200)]. Let me know if we should ask for `$country`.

Nyholm: ah, okey. I did not know

src/Provider/Geonames/Geonames.php
SrihariThalla: @Nyholm I'll fix all the styles at the end!

src/Provider/Geonames/Geonames.php
```diff
@@ -92,6 +98,77 @@ public function reverseQuery(ReverseQuery $query): Collection
...
+        }
+
+        $data = $json->geonames;
+
+        if (empty($data)) {
+            return new AddressCollection([]);
+        }
+
+        $results = [];
+        foreach ($data as $item) {
```
Nyholm: Will we ever have more than one item?

src/Provider/Geonames/Geonames.php
```diff
@@ -92,6 +98,77 @@ public function reverseQuery(ReverseQuery $query): Collection
...
+            $builder = new AddressBuilder($this->getName());
```
Nyholm: Do we really need the builder here?

Nyholm: We are not building an `Address` nor something that extends from the `Address`

src/Provider/Geonames/Geonames.php
```diff
@@ -92,6 +98,77 @@ public function reverseQuery(ReverseQuery $query): Collection
...
+
+            $builder->setBounds($item->south, $item->west, $item->north, $item->east);
```
Nyholm: `Bounds` does not exist on `CountryInfo`.

src/Provider/Geonames/Geonames.php
SrihariThalla: Yes, if the country codes requested are multiple (like `IN,US` instead of just `IN` or even if the country code is omitted altogether,) the result is multiple!

Sample: http://api.geonames.org/countryInfoJSON?lang=it&country=IN,US&username=demo&style=full
(Use my username if it reaches daily limit.)

Nyholm: Okey. Thank you.

src/Provider/Geonames/Geonames.php
SrihariThalla: Removed

src/Provider/Geonames/Geonames.php
SrihariThalla: `Bounds` is now set on `CountryInfo`

src/Provider/Geonames/Geonames.php
```diff
@@ -92,6 +98,73 @@ public function reverseQuery(ReverseQuery $query): Collection
...
 
     /**
+     * @param string|null $country
+     * @param string|null $locale
+     * @param int|null    $maxRows
+     * @param int|null    $startRow
+     * @return AddressCollection
+     * @throws \Geocoder\Exception\Exception
+     */
+    public function getCountryInfo(string $country = null, string $locale = null, int $maxRows = null, int $startRow = null): AddressCollection
```
SrihariThalla: `AddressCollection` requires a collection of `Location`s and `CountryInfo` is not a `Location`. Can you advise how should this be handled?

SrihariThalla: If `$country` is left out, then Geonames returns all the countries (250 items as of now). Let me know if this is fine or I will update it to a necessary param!

Nyholm:
Awesome. Looks good. Just add a test, changelog and make sue StyleCI is happy.

src/Provider/Geonames/Geonames.php
Nyholm: Can you just return an array or `CountryInfo`?

Nyholm: Maybe we should use better defaults?

`getCountryInfo(string $country = null, string $locale = null, int $maxRows = 10, int $startRow = 0)`

src/Provider/Geonames/Geonames.php
```diff
@@ -92,6 +98,75 @@ public function reverseQuery(ReverseQuery $query): Collection
...
+     * @param string|null $country
+     * @param string|null $locale
+     * @param int|null    $maxRows
+     * @param int|null    $startRow
+     *
+     * @return array
+     *
+     * @throws \Geocoder\Exception\Exception
+     */
+    public function getCountryInfo(string $country = null, string $locale = null, int $maxRows = 10, int $startRow = 0): array
```
SrihariThalla: @Nyholm It seems Geonames is not respecting `maxRows`. Shall we drop support for it?

http://api.geonames.org/countryInfoJSON?username=sriharithalla&style=full

Nyholm: I do not see startRow having any effect either...

Yeah, drop it =)

src/Provider/Geonames/Model/CountryInfo.php
```diff
@@ -0,0 +1,394 @@
...
+
+    /**
+     * @var string|null
+     */
+    private $currencyCode;
+
+    /**
+     * @return Bounds
+     */
+    public function getBounds(): Bounds
```
Nyholm: We cannot use `: Bounds` here. It may be null.

SrihariThalla:
@Nyholm All the requested changes are implemented. Please review when you can and let me know if any other changes are needed (any stashing of the commits)!

Nyholm:
I think this all looks good. Could you rebase on master to make sure travis is green before I merge?

SrihariThalla:
@Nyholm This should be it!

Nyholm:
Awesome. Thanks!

Issue #715, opened 2017-06-23
type "premise" missing in google-maps-provider


nigam214:
Issues in address parsing from Google Map API
https://github.com/geocoder-php/google-maps-provider

Missing Type: 
"types" : [ "premise" ]

Example:
https://maps.googleapis.com/maps/api/geocode/json?address=28.6105873,77.3885597

Nyholm:
Oh. I thought we did. Can you create a PR with a test and fix for this? Make the PR to this repo

svrdlic:
Earlier this month we added 'subpremise', not 'premise' :)

Nyholm:
Ah, okey. We should add premise as well.

svrdlic:
And I could fix this one also, if nobody replays

svrdlic:
Why dont we add all of AddressType parameters from google map replay..? https://developers.google.com/maps/documentation/geocoding/intro#Types

svrdlic:
Why dont we add all parameters from google map replay?
https://developers.google.com/maps/documentation/geocoding/intro#Types

Maybe we can add them to the dynamic array, that @Nyholm added instead of dynamic parameters?

Nyholm:
Feel free to do so =)

nigam214:
@svrdlic please go ahead. sorry, couldn't able to reply earlier.

nigam214:
we can consider adding "formatted_address" as well. in my project, I wanted to get the formatted address of given lat/long and encounter your library and these missing components. as @svrdlic suggested, adding all address parameters could be a better option.

svrdlic:
@nigam214  "formatted_address" is already added, I think. 
I hope that I'll be able to create PR until tomorrow.

nigam214:
@svrdlic saw the "formatted_address" in the code. thank you for pointing it out.

https://github.com/geocoder-php/google-maps-provider/blob/master/GoogleMaps.php#L264-L266

svrdlic:
@Nyholm I have a problem with tests. I created new test in src/Provider/GoogleMaps/Tests/GoogleMapsTest.php, and when I run it, it says that functions that I added to GoogleAddress.php doesn't exit. Is it possible that it looks for those functions in vendor/geocoder-php/google-maps-provider?

Nyholm:
Please submit a PR so we can see your work.

svrdlic:
ok

aditya-cherukuri:
[...]

PR #732, opened 2017-07-02
adding address_components parameters

svrdlic:
Added address components listed on https://developers.google.com/maps/documentation/geocoding/intro#Types
except  sublocality_level_#
issue #715

### DIFF
--- a/src/Provider/GoogleMaps/GoogleMaps.php
+++ b/src/Provider/GoogleMaps/GoogleMaps.php
@@ -300,6 +300,9 @@ final class GoogleMaps extends AbstractHttpProvider implements Provider
             if ($builder->hasValue('point_of_interest')) {
                 $address = $address->withPointOfInterest($builder->getValue('point_of_interest'));
             }
+            if ($builder->hasValue('establishment')) {
+                $address = $address->withEstablishment($builder->getValue('establishment'));
+            }
             $results[] = $address;
 
             if (count($results) >= $limit) {
@@ -366,6 +369,7 @@ final class GoogleMaps extends AbstractHttpProvider implements Provider
             case 'airport':
             case 'park':
             case 'point_of_interest':
+            case 'establishment':
                 $builder->setValue($type, $values->long_name);
                 break;
 
--- a/src/Provider/GoogleMaps/GoogleMaps.php
+++ b/src/Provider/GoogleMaps/GoogleMaps.php
@@ -264,8 +264,41 @@ final class GoogleMaps extends AbstractHttpProvider implements Provider
             if (isset($result->formatted_address)) {
                 $address = $address->withFormattedAddress($result->formatted_address);
             }
+            if ($builder->hasValue('streetAddress')) {
+                $address = $address->withStreetAddress($builder->getValue('streetAddress'));
+            }
+            if ($builder->hasValue('intersection')) {
+                $address = $address->withIntersection($builder->getValue('intersection'));
+            }
+            if ($builder->hasValue('political')) {
+                $address = $address->withPolitical($builder->getValue('political'));
+            }
+            if ($builder->hasValue('colloquialArea')) {
+                $address = $address->withColloquialArea($builder->getValue('colloquialArea'));
+            }
+            if ($builder->hasValue('ward')) {
+                $address = $address->withWard($builder->getValue('ward'));
+            }
+            if ($builder->hasValue('neighborhood')) {
+                $address = $address->withNeighborhood($builder->getValue('neighborhood'));
+            }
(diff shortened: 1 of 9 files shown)


src/Provider/GoogleMaps/GoogleMaps.php
```diff
@@ -321,8 +354,19 @@ private function updateAddressComponent(AddressBuilder $builder, string $type, $
...
+            case 'ward':
+            case 'neighborhood':
+            case 'premise':
             case 'subpremise':
-                $builder->setValue('subpremise', $values->long_name);
+            case 'natural_feature':
+            case 'airport':
+            case 'park':
+            case 'point_of_interest':
+                $builder->setValue((strpos($type, '_')) ? lcfirst(str_replace('_', '', ucwords($type, '_'))) : $type, $values->long_name);
```
Nyholm: Why do we need to transform it from snake_case to camelCase?

Couldn't we just do: 

```php
default:
  if (isset($values->long_name)) {
      $builder->setValue($type, $values->long_name);
  }
  break;
```

src/Provider/GoogleMaps/Model/GoogleAddress.php
```diff
@@ -103,7 +158,197 @@ public function withFormattedAddress(string $formattedAddress = null)
...
+        $new = clone $this;
+        $new->neighborhood = $neighborhood;
+
+        return $new;
+    }
+
+    /**
+     * @return null|string
+     */
+    public function getPark(): string
```
Nyholm: We cannot use `:string` here since "park" might return null. Same for all others.

svrdlic: I thought about it.. I'll remove it.

src/Provider/GoogleMaps/GoogleMaps.php
```diff
@@ -264,9 +264,42 @@ private function fetchUrl(string $url, string $locale = null, int $limit, string
             if (isset($result->formatted_address)) {
                 $address = $address->withFormattedAddress($result->formatted_address);
             }
+            if ($builder->hasValue('streetAddress')) {
```
svrdlic: @Nyholm  sorry for late response. I was thinking about replacing all this $builder->hasValue(..) with 
			if (!$builder->isEmptyDataArray())
			{
				$data = $builder->getDataArray();
				{
					foreach($data as $key => $value)
					{
						$address = $address->withData($key, $value);
					}
				}
			}


and 
public function withData(string $name, $value)
{
	$new = clone $this;
	$new->$name = $value;

	return $new;
}

Thats why I used camelCase in $data array. Should I do it, so we can discuss it?

Nyholm: If you do, that would not be type safe, right?

svrdlic: Hm.. All params are strings, why it wont be type safe?

Nyholm: Because you cannot do `$address->getPointOfInterest()`;

Also dynamic properties like `$new->$name = $value;` is bad for performance.

svrdlic: Ok, thanks for reply

svrdlic:
Sorry. I'll finish this tomorrow I hope..

src/Provider/GoogleMaps/GoogleMaps.php
```diff
@@ -264,9 +264,42 @@ private function fetchUrl(string $url, string $locale = null, int $limit, string
             if (isset($result->formatted_address)) {
                 $address = $address->withFormattedAddress($result->formatted_address);
             }
+            if ($builder->hasValue('street_address')) {
+                $address = $address->withStreetAddress($builder->getValue('street_address'));
+            }
```
Nyholm: Since all of the address setters allow null, can't we write this code more simple? Without the extra `if`?

`$address = $address->withStreetAddress($builder->getValue('street_address'));`

svrdlic: Actually, I couldn't find any example that has this parameter in address_component. Also I couldn't find example for `intersection`.

If we remove that `if` for `street_address`, why don't we remove it for all dynamic parameters.. But why should we return bunch of empty parameters?

Nyholm: Yes of course for all of these parameters. 

The result will be exactly the same and the code will get easier to read and far less complex (nPath)

svrdlic: Yes, response will be the same.. :)

PR #736, opened 2017-07-09
Removed GetLimit

### LINKED ISSUE #733: Remove some public methods from ProviderAggregator
Do we really need `getLimit`? Do we need the setter for limit?

Also, if #731 get merged, should we change the order of constructor arguments?

Are there any more public function we should remove in this class? (Or any other class?)

Nyholm:
This will fix #733


willdurand:
Is there any replacement for that? (sorry if it is a silly question.. )

Nyholm:
Thank you for reviewing this. There are no silly questions. 

I've removed the setter because we want all our providers and geocoders stateless. 
I removed the getter since we do not use it internally and because having a setter to read a static configuration value does not give much use.

PR #742, opened 2017-07-15
Adding Plugin support

Nyholm:
When creating the BazingaBundle we could either decorate all providers with a CacheProvider, LoggerProvider, ProfilingProvider etc... or we could use a PluginProfiler. This basically rips all nice plugin classes created by php-http and modify them to our needs. 

I put these plugin files in a new package, `geocoder-php/plugin`, because they have a few external dependencies. 

I've added 6 plugins to modify or log the queries. 

FYI @joelwurtz, @sagikazarmark

(Blocked by #741)

### Usage

```php
$provider = new GoogleMapsProvider();
$pluginProvider = new PluginProvider($provider, [
  new LocalePlugin('sv'),
  new CachePlugin($cacheService),
  new LoggerPlugin($logger),
]);

$query = GeocodeQuery::create('foobar');
$result = $pluginProvider->geocodeQuery($query);

// $result is in Swedish
// $result is cached
// $result is logged

// Try again
$result = $pluginProvider->geocodeQuery($query);

// $result is fetched from cache and not log were written. 

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,9 +33,9 @@
         "php-http/mock-client": "^1.0",
         "nyholm/psr7": "^0.2.2",
         "nyholm/nsa": "^1.1",
-        "cache/simple-cache-bridge": "^0.1.1",
-        "cache/array-adapter": "^0.5.0",
-        "cache/void-adapter": "^0.4.1"
+        "cache/simple-cache-bridge": "^1.0",
+        "cache/array-adapter": "^1.0",
+        "cache/void-adapter": "^1.0"
     },
     "suggest": {
         "ext-geoip": "Enabling the geoip extension allows you to use the MaxMindProvider.",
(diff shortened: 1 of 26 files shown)


willdurand:
I would reschedule this feature for version 4.1.0. I like the idea but not sure if it is the right approach. For instance,  we could have some features built-in in the core.

Nyholm:
Yeah, this is no blocker for 4.0.0. It only blocks the work on the bundle. 

> For instance, we could have some features built-in in the core.

What do you mean? We already have the StatefulGeocoder which adds bounds, locale, limit etc. We do also have a CacheProvider. The only **new** feature this bing is the method how we add things and the possibility to allow users and other library to add features.

arubacao:
👍 for Decorator Pattern

Nyholm:
You can see the need for this package in: https://github.com/geocoder-php/BazingaGeocoderBundle/issues/150

I want to move forward with this new package. Does anyone have an argument why this should not be created?

willdurand:
> I want to move forward with this new package. Does anyone have an argument why this should not be created?

Let's do it!

PR #743, opened 2017-07-16
Add withText and withCoordinates.

Nyholm:
Decorators or plugins may want to rewrite the query somehow. This PR allows that.

### DIFF
--- a/src/Common/Query/GeocodeQuery.php
+++ b/src/Common/Query/GeocodeQuery.php
@@ -71,6 +71,19 @@ final class GeocodeQuery
     }
 
     /**
+     * @param string $text
+     *
+     * @return GeocodeQuery
+     */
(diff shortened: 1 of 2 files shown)


willdurand:
👍

Nyholm:
Thank you for the review.

src/Common/Query/GeocodeQuery.php
```diff
@@ -71,6 +71,19 @@ public static function create(string $text): GeocodeQuery
     }
 
     /**
+     * @param string $text
+     *
+     * @return GeocodeQuery
+     */
+    public function withTest(string $text): GeocodeQuery
```
arubacao: there is a typo. `test` instead of `text`

Nyholm: Oh. Thank you!

PR #744, opened 2017-07-16
Added locale aware support for FreeGeoIp provider

shtricul:
Make support of locale aware for FreeGeoIp provider

The beginning of discussion is [here](https://github.com/geocoder-php/Geocoder/pull/479).

```php
$provider = new FreeGeoIp($adapter);

$provider->setLocale('fr');
$results = $provider->geocodeQuery(GeocodeQuery::create('81.27.51.251'));
/* or */
$results = $provider->geocodeQuery(GeocodeQuery::create('81.27.51.251')->withLocale('fr'));
```

### DIFF
--- a/src/Provider/FreeGeoIp/FreeGeoIp.php
+++ b/src/Provider/FreeGeoIp/FreeGeoIp.php
@@ -14,12 +14,12 @@ namespace Geocoder\Provider\FreeGeoIp;
 
 use Geocoder\Collection;
 use Geocoder\Exception\UnsupportedOperation;
-use Geocoder\Http\Provider\AbstractHttpProvider;
 use Geocoder\Model\AddressBuilder;
 use Geocoder\Model\AddressCollection;
-use Geocoder\Provider\Provider;
 use Geocoder\Query\GeocodeQuery;
 use Geocoder\Query\ReverseQuery;
+use Geocoder\Http\Provider\AbstractHttpProvider;
+use Geocoder\Provider\Provider;
 use Http\Client\HttpClient;
 use Psr\Http\Message\RequestInterface;
 
@@ -43,11 +43,7 @@ final class FreeGeoIp extends AbstractHttpProvider implements Provider
      * @param string     $locale
      * @param string     $baseUrl
      */
-    public function __construct(
-        HttpClient $client,
-        string $locale = '',
-        string $baseUrl = 'https://freegeoip.net/json/%s'
-    )
+    public function __construct(HttpClient $client, string $locale = '', string $baseUrl = 'https://freegeoip.net/json/%s')
     {
         parent::__construct($client);
 
--- a/src/Provider/FreeGeoIp/FreeGeoIp.php
+++ b/src/Provider/FreeGeoIp/FreeGeoIp.php
@@ -114,7 +114,7 @@ final class FreeGeoIp extends AbstractHttpProvider implements Provider
     /**
      * @return string|null
      */
-    public function getLocale(): ?string
+    public function getLocale()
     {
         return $this->locale;
     }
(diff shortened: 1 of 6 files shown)


src/Provider/FreeGeoIp/FreeGeoIp.php
```diff
@@ -33,13 +34,20 @@
...
+     */
+    private $locale;
+
+    /**
      * @param HttpClient $client
+     * @param string     $locale
      * @param string     $baseUrl
      */
-    public function __construct(HttpClient $client, string $baseUrl = 'https://freegeoip.net/json/%s')
+    public function __construct(HttpClient $client, string $locale = 'en', string $baseUrl = 'https://freegeoip.net/json/%s')
```
Nyholm: Is "en" a good default value? 

In `getRequest()` you do a check if it is null or not.

shtricul: I think it's better, then it will be string or null. Another way - set it with empty string, and check for if ($this->locale) in getRequest method. How do you think?

Nyholm: Why is that better? Not setting any locale is the current behavior.

shtricul: So I must by default set it with empty string, right?

Nyholm: Sure!

src/Provider/FreeGeoIp/FreeGeoIp.php
```diff
@@ -92,4 +100,20 @@ public function getName(): string
...
+    /**
+     * {@inheritdoc}
+     *
+     * @throws \InvalidArgumentException
+     */
+    protected function getRequest(string $url): RequestInterface
+    {
+        $request = parent::getRequest($url);
+
+        if ($this->locale !== null) {
```
Nyholm: Check if `empty($this->locale)` instead. 

https://github.com/geocoder-php/Geocoder/pull/744#discussion_r127926709

shtricul: Thx, will do it!

Nyholm:
Thank you. It all looks good. But what if I do something like: 

```
$provider = new FreeGeoip('xxx', 'ru');
$query = GeocoderQuery::create('foo');
$query->withLocale('en');
$result = $provider->geocodeQuery($query);
```

The $result should be in English and not Russian, right?

shtricul:
Yeah, you right. May be I should remove `locale` parameter from constructor and listen only for `locale` from query?

Nyholm:
I think it is good to have both. Then you can set one default (constructor) and one that temporally overrides it. 

```php
$provider = new FreeGeoip('xxx', 'ru');
$provider->geocodeQuery(GeocoderQuery::create('foo')); // Russian
$result = $provider->geocodeQuery(GeocoderQuery::create('foo')->withLocale('en')); // Englsh
$provider->geocodeQuery(GeocoderQuery::create('bar')); // Russian
```

Nyholm:
Ping

Nyholm:
I will close this PR after Christmas if nobody picks up the work.

Stadly:
Is this PR truly needed? Of course the code for setting the `Accept-Language` header for `FreeGeoIp` requests must be there, but that could use the value of `$query->getLocale()` instead of `$this->locale`. If one wants the locale to be used for all requests done by the provider instead of calling `withLocale()` on each request separately, one can either use `StatefulGeocoder` or `PluginProvider`, if I've understood everything correctly.

Nyholm:
Closing this PR due inactivity

Stadly:
Do you have any thoughts regarding my previous comment? If I've understood correctly, I would consider picking up the work on implementing locale support for FreeGeoIp.

Nyholm:
Sorry. 

I agree with your comment. See https://github.com/geocoder-php/Geocoder/pull/744#issuecomment-316303913

Stadly:
I've created a new PR: https://github.com/geocoder-php/Geocoder/pull/829

PR #748, opened 2017-07-24
Do not convert to UTF-8 if already UTF-8

ruicampos:
We're using GeoIPs provider with Maxmind binary database and we started getting weird enconding results. We found out that coming from database some strings are already UTF-8.
Running ``utf8_encode()`` on those strings will break the encoding.
This fix will only apply ``utf8_encode()`` if the string is not UTF-8

Tests are green.

### DIFF
--- a/src/Geocoder/Provider/AbstractProvider.php
+++ b/src/Geocoder/Provider/AbstractProvider.php
@@ -115,7 +115,7 @@ abstract class AbstractProvider
     protected function fixEncoding(array $results)
     {
         return array_map(function ($value) {
-            return is_string($value) ? utf8_encode($value) : $value;
+            return is_string($value) && mb_detect_encoding($value) != 'UTF-8' ? utf8_encode($value) : $value;
         }, $results);
     }
 }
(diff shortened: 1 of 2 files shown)


Nyholm:
Also, you targeted an old branch and not master.

ruicampos:
Changes:
* Strict comparison
* Test to prove it

I targeted 3.x branch as we're using 3.* version and found the problem. 

I quickly looked into master (version 4) and it seems it doesn't have the double encoding problem as it is not fixing encodings on GeoIP2 - so haven't done anything on master.

ping @Nyholm

Nyholm:
Okey. Thank you. I'll be happy to merge if the style and Travis is green.

ruicampos:
Ok, "Boy Scout rule" then. :-)

Fixed the style. 

Tests are all ok except for hhvm which seems it is no longer supported by Travis:

.travis.yml

Sincerely... is anyone using hhvm these days? 
I can remove the line from [.travis.yml](https://github.com/geocoder-php/Geocoder/blob/3.x/.travis.yml) in this PR if you agree.

ping @Nyholm

Nyholm:
No that's fine if just hhvm is failing. I'll review the PR tomorrow. 

//Tobias Nyholm

Skickat på språng

Issue #749, opened 2017-07-24
ID from Service Provider in Address Model


arubacao:
Is there any chance on adding the original IDs from the service providers to the address model?
I'm heavily relying on them and have to keep an almost identical fork of some providers since inheriting the final base classes is not possible.

Cheers,
Chris

Nyholm:
We have the `Location::getProvidedBy`([source](https://github.com/geocoder-php/Geocoder/blob/master/src/Common/Location.php#L115)) which is the same as `Provider::getName`. Isn't that what you are asking for?

arubacao:
yes, I'm also using this method to distinguish between providers. But what I'm referring to in my question is the actual ID.
So somewhere within the [`updateAddressComponent`](https://github.com/geocoder-php/google-maps-provider/blob/master/GoogleMaps.php#L244) method I'd get the value from `place_id` and set it on [address object](https://github.com/geocoder-php/google-maps-provider/blob/master/GoogleMaps.php#L204)

> place_id is a unique identifier that can be used with other Google APIs. For example, you can use the place_id in a Google Places API request to get details of a local business, such as phone number, opening hours, user reviews, and more. See the [place ID overview](https://developers.google.com/places/place-id).

Nyholm:
Ah, on the GoogleMaps provider. Of course, we should definitely extend Google Maps' address object with a place ID. 

I would be happy to review and merge such PR.

arubacao:
Consider it done

arubacao:
Do you think it would make sense to add the `ID` to the `Geocoder\Model\Address` class?
I guess most (maybe all ?) providers provide some sort of identification.

PR #750, opened 2017-07-24
Add place_id to GoogleAddress

arubacao:
For #749 

Cheers,
Chris

### DIFF
--- a/src/Provider/GoogleMaps/Model/GoogleAddress.php
+++ b/src/Provider/GoogleMaps/Model/GoogleAddress.php
@@ -22,6 +22,11 @@ final class GoogleAddress extends Address
     /**
      * @var string|null
      */
+    private $id;
+
+    /**
+     * @var string|null
+     */
     private $locationType;
 
     /**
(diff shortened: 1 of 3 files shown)


src/Provider/GoogleMaps/Model/GoogleAddress.php
```diff
@@ -22,6 +22,11 @@
     /**
      * @var string|null
      */
+    private $id;
```
Nyholm: Maybe add a comment explaining that this is the "place id"

src/Provider/GoogleMaps/Model/GoogleAddress.php
```diff
@@ -100,6 +105,29 @@
...
+    {
+        $new = clone $this;
+        $new->id = $id;
+
+        return $new;
+    }
+
+    /**
+     * @see https://developers.google.com/places/place-id
+     *
```
Nyholm: Oh. This is great.

Issue #716, opened 2017-06-23
type "sublocality_level_#" missing in google-maps-provider


nigam214:
Issues in address parsing from Google Map API
https://github.com/geocoder-php/google-maps-provider

Missing Type: 
"types" : [ "political", "sublocality", "sublocality_level_1" ]
"types" : [ "political", "sublocality", "sublocality_level_2" ]

Example:
https://maps.googleapis.com/maps/api/geocode/json?address=28.6105873,77.3885597

Nyholm:
Thank you for both issues. Can you try to fix this as well? Maybe with the same PR if it makes sense?

svrdlic:
I could fix this issue, if nobody replays

Nyholm:
That would be great.

nigam214:
@svrdlic please fix it if you can. it will be very helpful.

Nyholm:
Closed by #751

PR #751, opened 2017-07-31
adding sublocality level for GoogleMap

svrdlic:
New pull request for issue #716
please review again @Nyholm, and I'm sorry about this mess


svrdlic:
I added new function in AddressBuilder, thats why travis has error. It already happened once before..

src/Common/Model/AddressBuilder.php
```diff
@@ -300,6 +300,19 @@ public function setValue(string $name, $value): self
     }
 
     /**
+     * @param string $name
+     * @param mixed  $value
+     *
+     * @return AddressBuilder
+     */
+    public function setValueInArray(string $name, $value): self
```
Nyholm: Do we really need this function?

src/Provider/GoogleMaps/GoogleMaps.php
```diff
@@ -267,6 +269,19 @@ private function updateAddressComponent(AddressBuilder $builder, string $type, $
...
+            case 'sublocality_level_2':
+            case 'sublocality_level_3':
+            case 'sublocality_level_4':
+            case 'sublocality_level_5':
+                $newSubLocalityLevel = [
+                    'level' => intval(substr($type, -1)),
+                    'name' => $values->long_name,
+                    'code' => $values->short_name,
+                ];
+                $builder->setValueInArray('subLocalityLevel', $newSubLocalityLevel);
```
Nyholm: What if we do not add it to the builder here but after the switch statement

src/Provider/GoogleMaps/GoogleMaps.php
```diff
@@ -267,6 +269,19 @@ private function updateAddressComponent(AddressBuilder $builder, string $type, $
                 $builder->addAdminLevel(intval(substr($type, -1)), $values->long_name, $values->short_name);
                 break;
 
+            case 'sublocality_level_1':
+            case 'sublocality_level_2':
+            case 'sublocality_level_3':
+            case 'sublocality_level_4':
+            case 'sublocality_level_5':
+                $newSubLocalityLevel = [
```
Nyholm: Let's do $newSubLocalityLevel[] = ...

src/Provider/GoogleMaps/Model/SubLocalityLevel.php
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @license    MIT License
+ */
+
+namespace Geocoder\Provider\GoogleMaps\Model;
+
+/**
+ * Class SubLocalityLevel is used only for GoogleMap provider and contains functions for working with concrete subLocalityLevel
+ */
+final class SubLocalityLevel
```
Nyholm: This class is exactly the same as admin level. Maybe we do not need to create a new class

src/Provider/GoogleMaps/Model/SubLocalityLevelCollection.php
```diff
@@ -0,0 +1,146 @@
...
+namespace Geocoder\Provider\GoogleMaps\Model;
+
+use Geocoder\Exception\CollectionIsEmpty;
+use Geocoder\Exception\InvalidArgument;
+use Geocoder\Exception\OutOfBounds;
+
+/**
+ * Class SubLocalityLevelCollection is used only for GoogleMap provider, contains functions for working with arrays of SubLocalityLevel
+ */
+final class SubLocalityLevelCollection implements \IteratorAggregate, \Countable
```
Nyholm: This looks exactly the same as AdminLevelCollection. Maybe we should use that instead

src/Provider/GoogleMaps/GoogleMaps.php
svrdlic: array $newSubLocalityLevel in defined in function, it has local scope.

src/Provider/GoogleMaps/Model/SubLocalityLevel.php
svrdlic: My first idea was to use `AdminLevel` & `AdminLevelCollection`. I actually make it that way, but  `AdminLevel` class has hardcoded word `adminLevel`. I dont think it is ok to have subLocalityLevel called adminLevel in the result.. So I decided to make new classes.

Nyholm: A "subLocallity level" actually is an administrative level. I think it is fine

src/Provider/GoogleMaps/GoogleMaps.php
Nyholm: I know. But add it to the builder after the switch statement. Not outside this function

svrdlic: Im sorry, but I dont see the difference if I add it after the switch statement. We call this function foreach address component type, and each time it will be initialized again, and overrides previous values.. Could you please explain me this again?

Nyholm: Something like this: 

```php
function updateAddressComponent() {
   $newSubLocalityLevel = [];
   switch ($type) {
      // ...
      case 'sublocality_level_1':
      case 'sublocality_level_2':
      case 'sublocality_level_3':
      case 'sublocality_level_4':
      case 'sublocality_level_5':
           $newSubLocalityLevel[] = [
                  'level' => intval(substr($type, -1)),
                  'name' => $values->long_name,
                  'code' => $values->short_name,
            ];
    } // End of switch
    $builder->setValue('subLocalityLevel', $newSubLocalityLevel);
}
```

With the code above you do not need the `setValueInArray` function.

svrdlic: I can do that, but Im affraid it wont work in the way we want.
We call updateAddressComponent(...) like this
```

            foreach ($result->address_components as $component) {
                foreach ($component->types as $type) {
                    $this->updateAddressComponent($builder, $type, $component);
                }
            }
```
and structure of result is  like this

```
{
   "results" : [
      {
         "address_components" : [
            {
               "long_name" : "58",
               "short_name" : "58",
               "types" : [ "political", "sublocality", "sublocality_level_4" ]
            },
            {
               "long_name" : "Iijima",
               "short_name" : "Iijima",
               "types" : [ "political", "sublocality", "sublocality_level_1" ]
            }
  ]
}
```
So, without new function it will work only if we do something like
`$builder->setValue('subLocalityLevel_#', $newSubLocalityLevel);
`
where # will be level's number. It that case I wouldnt have all sublocality levels in one  subarray.

src/Provider/GoogleMaps/Model/SubLocalityLevel.php
svrdlic: I changed it

src/Provider/GoogleMaps/Model/SubLocalityLevelCollection.php
svrdlic: Sure

src/Provider/GoogleMaps/GoogleMaps.php
svrdlic: I had similar idea, but $newSubLocalityLevel[] will make array with key 0 each time
`$newSubLocalityLevel = [0=> [ 
                                                 'level' => 3,
                                                 'name' => 'name'
                                                 'code' => 'code'
                                                 ]
                                         ]`

and `setValue(..)` will pass this array, and replace previous value (which also had 0 key). 
If we call  `setValue(..)`  after the switch statement like that, we would override `subLocalityLevel` subarray each time we call the function. 
Maybe I missed out on something?

Nyholm: Yeah. You only call setValue once for sublocallity.

svrdlic: We have subLocality and subLocalityLevels. I thought that we need all subLocalityLevels in our result, just like we have all adminLevels? 
```
{
   "results" : [
      {
         "address_components" : [
            {
               "long_name" : "58",
               "short_name" : "58",
               "types" : [ "political", "sublocality", "sublocality_level_4" ]
            },
            {
               "long_name" : "Iijima",
               "short_name" : "Iijima",
               "types" : [ "political", "sublocality", "sublocality_level_1" ]
            }
  ]
}
```

Nyholm: I meant subLocalityLevels. 

Why don't you build a local array with subLocalityLevels and then give that array to the builder? I do not see the issue with that.

svrdlic: Im sorry, I dont know how to to that. 

This function `updateAddressComponent()` is called foreach `types`. That means it will be called multiple times for each `address_components`, and each time array `$newSubLocalityLevel ` will be reset, and it will always contains just one, current, `subLocalityLevel`. 

And function `setValue(..)` will always reassign `$data['subLocalityLevel']`, and set it to the current `subLocalityLevel`.

```
   foreach ($result->address_components as $component) {
       foreach ($component->types as $type) {
               $this->updateAddressComponent($builder, $type, $component);
        }
   }
```
and here is result

```
{
   "results" : [
      {
         "address_components" : [
            {
               "long_name" : "58",
               "short_name" : "58",
               "types" : [ "political", "sublocality", "sublocality_level_4" ]
            },
            {
               "long_name" : "Iijima",
               "short_name" : "Iijima",
               "types" : [ "political", "sublocality", "sublocality_level_1" ]
            }
  ]
}

```

Nyholm:
I sent a PR to your fork.

Nyholm:
Tank you.

PR #773, opened 2017-09-12
In Nominatim provider, use city, then town, then village, then hamlet

SebastienTainon:
When using Nominatim provider, the API can return several kinds of responses for "city" for the same kind of query.
For example:

**75004, France**
https://nominatim.openstreetmap.org/search?q=75004,France&format=xml&addressdetails=1&limit=5

    <suburb>Quartier Saint-Merri</suburb>
    <city_district>4th Arrondissement</city_district>
    <city>Paris</city>
    <county>Paris</county>
    <state>Ile-de-France</state>
    <country>France</country>
    <postcode>75004</postcode>

=> we receive the city in a "city" field

**98000, Monaco**
https://nominatim.openstreetmap.org/search?q=98000%2C+Monaco&format=xml&addressdetails=1&limit=5

    <suburb>La Condamine</suburb>
    <town>Monaco</town>
    <postcode>98000</postcode>
    <country>Monaco</country>

=> we receive the city field in a "town" field

Here is the wiki for Nominatim structure: http://wiki.openstreetmap.org/wiki/Nominatim

**My proposition is that if the "city" field is absent from the response, we use the "town" field instead to define the city of the response.**

### DIFF
--- a/src/Provider/GoogleMaps/Model/GoogleAddress.php
+++ b/src/Provider/GoogleMaps/Model/GoogleAddress.php
@@ -116,7 +116,7 @@ final class GoogleAddress extends Address
      *
      * @return GoogleAddress
      */
-    public function withId(string $id = null): GoogleAddress
+    public function withId(string $id = null): self
     {
         $new = clone $this;
         $new->id = $id;
@@ -139,7 +139,7 @@ final class GoogleAddress extends Address
      *
      * @return GoogleAddress
      */
-    public function withLocationType(string $locationType = null): GoogleAddress
+    public function withLocationType(string $locationType = null): self
     {
         $new = clone $this;
         $new->locationType = $locationType;
@@ -168,7 +168,7 @@ final class GoogleAddress extends Address
      *
      * @return GoogleAddress
      */
-    public function withResultType(array $resultType): GoogleAddress
+    public function withResultType(array $resultType): self
     {
         $new = clone $this;
         $new->resultType = $resultType;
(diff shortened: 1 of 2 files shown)


SrihariThalla:
Just to extend the hierarchy, the locality can be returned as `village` and `hamlet` as well! Probably adding them as well might be good.

SebastienTainon:
@SrihariThalla Do you have an example of a Nominatim URL returning an `hamlet`? It does not appear in the documentation.
What about checking first city, then town, and then village if the two previous ones do not exist?

SrihariThalla:
Here are the links, @SebastienTainon :

- Village: https://nominatim.openstreetmap.org/search.php?q=Peddakurapadu&polygon_geojson=0&addressdetails=1&format=json
- Hamlet: https://nominatim.openstreetmap.org/search.php?q=Yerragondapalem&polygon_geojson=0&addressdetails=1&format=json

The preferred order would be like you said:

1. `city`
2. `town`
3. `village`
4. `hamlet`

SebastienTainon:
@SrihariThalla I've updated the PR with tht modifications you are suggesting. Could you check that it's ok for you? :smile:

Nyholm:
@SebastienTainon could you rebase on master so I can merge?

SebastienTainon:
I've rebased and squashed @Nyholm :)

Nyholm:
Can you also fix (the new) style CI stuff?

SebastienTainon:
It was related to #789 but no problem, I've fixed it too ;)

SrihariThalla:
Sorry, I lost track of this :-/ All Nominatim changes look good 👍

PR #789, opened 2017-10-10
GoogleMaps: Channel parameter

dkemper:
You can add an optional parameter to the google maps requests to see the usage on a specific channel.

I added this parameter in this PR.
I also made some code style fixes. I hope its useful for you.

The description is here.
https://developers.google.com/maps/premium/overview?hl=en

### DIFF
--- a/src/Provider/GoogleMaps/Model/GoogleAddress.php
+++ b/src/Provider/GoogleMaps/Model/GoogleAddress.php
@@ -116,7 +116,7 @@ final class GoogleAddress extends Address
      *
      * @return GoogleAddress
      */
-    public function withId(string $id = null) : GoogleAddress
+    public function withId(string $id = null): GoogleAddress
     {
         $new = clone $this;
         $new->id = $id;
@@ -139,7 +139,7 @@ final class GoogleAddress extends Address
      *
      * @return GoogleAddress
      */
-    public function withLocationType(string $locationType = null) : GoogleAddress
+    public function withLocationType(string $locationType = null): GoogleAddress
     {
         $new = clone $this;
         $new->locationType = $locationType;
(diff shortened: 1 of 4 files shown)


src/Provider/GoogleMaps/Model/GoogleAddress.php
```diff
@@ -483,15 +483,15 @@ public function getSubLocalityLevels()
...
+    public function withSubLocalityLevels(array $subLocalityLevel): GoogleAddress
     {
         $subLocalityLevels = [];
         foreach ($subLocalityLevel as $level) {
             if (empty($level['level'])) {
                 continue;
             }
 
-            $name = $level['name'] ?? $level['code'] ?? null;
+            $name = $level['name'] ?? $level['code'] ?? '';
```
Nyholm: Why was this line changed?

src/Provider/GoogleMaps/Tests/GoogleMapsTest.php
```diff
@@ -282,6 +284,33 @@ function (RequestInterface $request) use (&$uri) {
...
+            null,
+            null,
+            'bar'
+        );
+
+        try {
+            $provider->geocodeQuery(GeocodeQuery::create('blah'));
+        } catch (InvalidServerResponse $e) {
+        }
+        $this->assertEquals(
```
Nyholm: Maybe do a `asserRegex` instead. This test is likely to change when we add a new parameter.

src/Provider/GoogleMaps/Model/GoogleAddress.php
dkemper: new AdminLevel() $name parameter requires a string. Its never called because of the empty check before, but the given name parameter should be a string. 
It doesnt matter, but my editor shows it as wrong.

Nyholm: Great. Thank you.

src/Provider/GoogleMaps/Tests/GoogleMapsTest.php
dkemper: I just copied it from the other test. I can replace both if you want.

PR #791, opened 2017-10-27
Add GoogleMaps Missing Fields

ConnorVG:
This adds `establishment` and `point_of_interest` which are useful as the `streetNumber` when the address is not a numbered address (such as Disney World).

### DIFF
--- a/src/Geocoder/Provider/GoogleMaps.php
+++ b/src/Geocoder/Provider/GoogleMaps.php
@@ -250,6 +250,8 @@ class GoogleMaps extends AbstractHttpProvider implements LocaleAwareProvider
                 $resultSet['countryCode'] = $values->short_name;
                 break;
 
+            case 'establishment':
+            case 'point_of_interest':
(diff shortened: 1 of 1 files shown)


Nyholm:
Checkout the latest version of the Geocoder. These fields are already added: https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/GoogleMaps/GoogleMaps.php#L231

ConnorVG:
@Nyholm and I love that they are, the problem is that we're stuck with version 3.3.x and this seems like a feature it should sensibly support

Nyholm:
3.3 was feature frozen in January and will reach EOL for bugfixes next week. See https://github.com/geocoder-php/Geocoder#3x

ConnorVG:
Thank you, I understand. I'll see what I can do in terms of moving to 4.

Nyholm:
Awesome 👍 

Please write an issue if there is anything weird you find during migration.

PR #798, opened 2017-11-21
Added small clarifications for string formatter.

leevigraham:

PR #799, opened 2017-11-22
StyleCi fixes

leevigraham:
As reported: https://styleci.io/analyses/qxKNJB in PR: https://github.com/geocoder-php/Geocoder/pull/798


Nyholm:
Thank you!

PR #806, opened 2017-12-29
Improve Nominatim Provider

### LINKED ISSUE #805: Nominatim Provider improvements
Hello everyone (and happy holidays 🎉 ),

I started to have a look at `geocoder-php` and it's really awesome ! 👍 
I use *Nominatim* a lot and since now, I did my query directly on the Nominatim API.

I feel like there are useful items missing in the Nominatim Provider :
- The **locationType** provided by `class` parameter ;
- The **formattedAddress** provided by `display_name` parameter ;
- The **copyright**/**licence** (indeed, the [Nominatim Usage Policy](https://operations.osmfoundation.org/policies/nominatim/) asks to display clearly the attributions) provided by `attribution` parameter ;

Since, it's my first use of `geocoder-php` I would like to know how you feel about this before suggesting any PR and what would be the best way to do it ? Should I base Nominatim Provider on Google Maps provider for instance (which includes **locationType** and **formattedAddress**) ?

Thanks a lot !

jbelien:
Add following fields:
- Attribution
- Class
- Display name
- OSM type
- OSM Id
- Type

Close #805


Nyholm:
Thank you. After a first review I think it looks good. 

Could you fix the style errors and add some tests?

jbelien:
> Could you fix the style errors and add some tests?

Doing it right now :)

jbelien:
Style errors are fixed.

I'm quite new with PHP Unit so I might take some time to write the tests ; I'll do my best :)

Nyholm:
We just need some integration testing. Ie, write a test and verify the result. The HTTP response will be cached, so make sure to commit the cached file. 

It is really just copy paste from existing tests. 

I want a test that shows the result has a class, display name, attribution etc.

jbelien:
Okay, I'll try to figure this out :P

jbelien:
I think it's done :)
Tell me if I forgot something.

src/Provider/Nominatim/Nominatim.php
```diff
@@ -169,7 +174,23 @@ private function xmlResultToArray(\DOMElement $resultNode, \DOMElement $addressN
             $builder->setBounds($bounds['south'], $bounds['west'], $bounds['north'], $bounds['east']);
         }
 
-        return $builder->build();
+        if (false === $reverse) {
```
Nyholm: Do we really need this separation? It seams like most the lines are the same. 
I suggest to remove the `$reverse` parameter.

src/Provider/Nominatim/Nominatim.php
```diff
@@ -160,6 +164,7 @@ private function xmlResultToArray(\DOMElement $resultNode, \DOMElement $addressN
             $countryCode = strtoupper($countryCode);
         }
         $builder->setCountryCode($countryCode);
+
```
Nyholm: Why what this blank line added?

src/Provider/Nominatim/Nominatim.php
```diff
@@ -149,7 +154,6 @@ private function xmlResultToArray(\DOMElement $resultNode, \DOMElement $addressN
             }
         }
 
-        $builder->setPostalCode($postalCode);
```
Nyholm: Why was this moved?

src/Provider/Nominatim/Nominatim.php
```diff
@@ -123,7 +127,7 @@ public function reverseQuery(ReverseQuery $query): Collection
      *
      * @return Location
      */
-    private function xmlResultToArray(\DOMElement $resultNode, \DOMElement $addressNode): Location
+    private function xmlResultToArray(\DOMElement $resultNode, \DOMElement $addressNode, string $attribution = 'Data © OpenStreetMap contributors, ODbL 1.0. http://www.openstreetmap.org/copyright', bool $reverse): Location
```
Nyholm: Will the default value ever have an effect here? 

See https://3v4l.org/F1LKl

Let's remove the default value.

src/Provider/Nominatim/Model/NominatimAddress.php
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * @return null|int
+     */
+    public function getOSMId()
+    {
+        return $this->osmId;
+    }
+
+    /**
+     * @param null|string $osmId
```
Nyholm: int?

jbelien: Indeed, I'll fix it !

src/Provider/Nominatim/Nominatim.php
jbelien: 👍

src/Provider/Nominatim/Nominatim.php
jbelien: Seemed more logical to regroup postal code process ;
But I really don't mind putting it back to line 152 ^_^

src/Provider/Nominatim/Nominatim.php
jbelien: Same as previous comment, seemd to make more sense to create a "block" with country code process ;
But again, really don't mind putting that back as it was :)

src/Provider/Nominatim/Nominatim.php
jbelien: Yes, I don't really like that solution with `reverse` parameter either but the API answer from geocoding or reverse geocoding is quite different so the extraction of the files is also different :

**Geocoding:**
```$location = $location->withDisplayName($resultNode->getAttribute('display_name'));```
**Reverse geocoding:**
```$location = $location->withDisplayName($resultNode->nodeValue);```

Any suggestion to improve/simplify that ?

jbelien:
Happy new year ! 🎉

src/Provider/Nominatim/Nominatim.php
Nyholm: I see.. Lets improve it by only putting the differences in the if-else-statement. It is a private function so we may improve it later if we come up with anything.

src/Provider/Nominatim/Nominatim.php
Nyholm: Okey. I just saw a weird diff. This is fine

src/Provider/Nominatim/Nominatim.php
jbelien: 👍 I'll take care of that today !

jbelien:
Done ! :)

PR #813, opened 2018-01-26
Return empty collection if address was not found

Stadly:
If the location of the ip address cannot be determined by FreeGeoIp, return an empty AddressCollection instead of dummy data. This is particularly useful when using the Chain provider.


Stadly:
I'm don't understand why the Travis CI build fails. Could you point me in the right direction for fixing it?

Nyholm:
Build failed because of temporary Travis issue. 

Thank you for this PR.

PR #817, opened 2018-02-09
Add Coordinates::toArray

Mechazawa:


Nyholm:
Could you give me some more background why we should have this?

Mechazawa:
For our use-case we need to get the coordinates as a tuple. Instead of having to do this every time a nice helper method would be nice. 

```php
$geoCoordinates = $address->getCoordinates();
$coordinates = [$geoCoordinates->getLatitude(), $geoCoordinates->getLongitude()];
```

jbelien:
I like the `toArray()` idea **but** I would put **`[lon, lat]`** instead of `[lat, lon]` to respect standards ; see [GeoJSON standard](https://tools.ietf.org/html/rfc7946#appendix-A) for instance : 
> Point coordinates are in x, y order (easting, northing for projected coordinates, longitude, and latitude for geographic coordinates)

Mechazawa:
I've flipped the response tuple

PR #820, opened 2018-02-09
Removed return types

Nyholm:
I would love to keep this, but it is a BC break. :/

### DIFF
--- a/src/Provider/GoogleMaps/Model/GoogleAddress.php
+++ b/src/Provider/GoogleMaps/Model/GoogleAddress.php
@@ -116,7 +116,7 @@ final class GoogleAddress extends Address
      *
      * @return GoogleAddress
      */
-    public function withId(string $id = null): self
+    public function withId(string $id = null)
     {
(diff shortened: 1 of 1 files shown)


src/Provider/GoogleMaps/Model/GoogleAddress.php
```diff
@@ -116,7 +116,7 @@
      *
      * @return GoogleAddress
      */
-    public function withId(string $id = null): self
```
mbrodala: Technically this is not really accurate anways. It should be `GoogleAddress` instead since it's not the same instance that is returned when calling these methods.

Nyholm: Writing `:GoogleAddress` or `:self` is the same. `:self` means the same class. Not the same object.

mbrodala: OK, you are right. So far I had in mind that this was the native `@return $this` solution.

PR #821, opened 2018-02-09
Added changelog for php-common

Nyholm:


Nyholm:
Before 4.2 we should consider: https://github.com/thephpleague/geotools/pull/109

PR #833, opened 2018-03-02
Add Query::withText

mbrodala:
This was missing in #743 and is essential for mocking in tests.

### DIFF
--- a/src/Common/Query/Query.php
+++ b/src/Common/Query/Query.php
@@ -23,7 +23,7 @@ interface Query
      * @return Query
      */
     public function withText(string $text);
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you for this PR. I cannot remember the reason why that function was not a part of the interface. Maybe we considered it unnecessary. 

I cannot merge this PR now since it would be a BC break. I will put it to the v5 milestone. 
(But do not hold your breath about v5 to be released... sorry)

mbrodala:
Sure, I understand. I'll try to figure out a working test setup.

PR #837, opened 2018-03-15
Make ProviderCache not final

cookieguru:
I'd like my cache keys to be namespaced.  The easiest way for me to do so would be to simply override the `getCacheKey` method to work how I want and pass that to the geocoder.  However that's not possible with it being final.
I don't want to reimplement every single method (read: copy and paste).  Therefore I propose marking this class _not_ final so that this method and the `getName` method can be overridden.  I have marked the other methods as final and changed the private properties to protected.

### DIFF
--- a/src/Provider/Cache/CHANGELOG.md
+++ b/src/Provider/Cache/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+## 4.1.0 (2018-03-16)
+
+* Changed: `ProviderCache` is no longer final allowing `ProviderCache::getCacheKey` to be overridden
+
 ## 4.0.1
 
 ### Changed
--- a/src/Provider/Cache/ProviderCache.php
+++ b/src/Provider/Cache/ProviderCache.php
@@ -26,26 +26,26 @@ final class ProviderCache implements Provider
     /**
      * @var Provider
      */
-    private $realProvider;
+    protected $realProvider;
 
     /**
      * @var CacheInterface
      */
-    private $cache;
+    protected $cache;
 
     /**
      * How log a result is going to be cached.
      *
      * @var int|null
      */
-    private $lifetime;
+    protected $lifetime;
 
     /**
      * @param Provider       $realProvider
      * @param CacheInterface $cache
      * @param int            $lifetime
      */
-    public function __construct(Provider $realProvider, CacheInterface $cache, int $lifetime = null)
+    final public function __construct(Provider $realProvider, CacheInterface $cache, int $lifetime = null)
     {
         $this->realProvider = $realProvider;
         $this->cache = $cache;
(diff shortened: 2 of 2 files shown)


Nyholm:
Could you write a change log in this provider for a new minor release?

cookieguru:
Sure, would this go under a new 4.0.1 release in `CHANGELOG.md` or...?

Nyholm:
You should put it under a new heading called 4.1.0 in this file: https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/Cache/CHANGELOG.md.

cookieguru:
Done

Nyholm:
Thanks. 
But I think you updated the wrong changelog. 
The correct one is: [/src/Provider/Cache/CHANGELOG.md](https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/Cache/CHANGELOG.md)

cookieguru:
Oops, you're right.  Fixed now

Nyholm:
Thanks

Nyholm:
Oops. I see that we both missed to remove the `final` keyword on the class =)

unkind:
IMHO, it would be better to pass 4th argument like `callable $cacheKeyStrategy = null`.

cookieguru:
@unkind I disagree, dependency injection is not the right way to solve this.  By using that method you have no access to the cache instance itself, which may itself have some properties or methods that are necessary to compute the cache key.  Your method also forces the re-declaration of the method upon instantiation, which isn't always necessary or desired.

unkind:
@cookieguru 

> have no access to the cache instance itself

Do you mean 

```php
new ProviderCache($realProvider, $cache, null, function (string $query) use ($cache) {
    // namespace logic for keys and you have $cache instance here
});
```
?

> which may itself have some properties or methods that are necessary to compute the cache key

Can you provide the use case?

> Your method also forces the re-declaration of the method upon instantiation, which isn't always necessary or desired.

Even if it is a problem, you may use lazy loading. Your `protected` method in fact is lazy loaded.

cookieguru:
> You can use DI-factories to instantiate just once.

What I meant is that re-declaring the method is *required* even if you are agnostic about its implementation.

> It's a weird to inherit something in order to substitute something

I disagree.  That's the whole reason why we have abstract classes.


If you look at `CacheProvider` it is itself just a decorator of `Provider`.  Your criticism of overwriting this method alone seems inconsistent.

unkind:
> What I meant is that re-declaring the method is required even if you are agnostic about its implementation.

No. You can pass `null` for default strategy.

> That's the whole reason why we have abstract classes.

`ProviderCache` is not abstract class and `final` was pointing on it very clearly.

> If you look at CacheProvider it is itself just a decorator of Provider.

It is implementation of `Provider`. Implement interface is find, inherit concrete class is a bad sign.

cookieguru:
May I suggest [this](https://github.com/geocoder-php/Geocoder/fork) if you're unhappy with everyone else's implementation?

unkind:
> May I suggest this if you're unhappy with everyone else's implementation?

Does it mean you can't speak reasonably anymore?

Issue #839, opened 2018-03-21
Use Address object as query, instead of just one line of free text


TerjeBr:
You have this class `Geocoder\Model\Address` that implements `Geocoder\Location`. But is is only used when returning responses/results.

What I would like to see is a way to also pass this as the request to the geocoder. Currently it is only possible to pass the address in the request as free formatted text string.

The background for this is that I noticed that at mapquest.com you can pass something they call a 5-box format to give better results. See [Map Quest API](https://developer.mapquest.com/documentation/open/geocoding-api/) and [Specifying Locations](https://developer.mapquest.com/documentation/common/forming-locations/).

One way around this may be to pass the Address object on the data section of the query. But I thought that seemed a bit clunky. Since the only purpose of this library is to send queries about addresses, should it not be a more streamlined way to send an actual `Address` object as a query instead of just a free text string?

Nyholm:
Thank you for this issue. You could pass the GeocodeQuery some arbitrary data. See https://github.com/geocoder-php/Geocoder/blob/master/src/Common/Query/GeocodeQuery.php#L131

I would love to see a PR to MapQuest that supports this.

TerjeBr:
Yes, I know I can use `GeocodeQuery->withData()` to send arbitary data along with the query.

But my question was if you would consider changing the `GeocodeQuery` class, along with the interface `Query` to allow for a more direct or explicit reference to an object implementing `Geocoder\Location`?

Then various providers could may be call `$query->getLocation()` as well as `$query->getText()`?

As I said, it seems a bit clunky that the standard way to pass an `Address` object should be in the data section of the query.

TerjeBr:
May be a solution is to just extend `GeocodeQuery` and send that. If I make a PR where would that extended `GeocodeQuery` live? In the MapQuest provider directory, or in the Common directory?

Nyholm:
> But my question was if you would consider changing the GeocodeQuery class, along with the interface Query to allow for a more direct or explicit reference to an object implementing Geocoder\Location?

No, that would be a BC break. 

> Then various providers could may be call $query->getLocation() as well as $query->getText()?

Right now there is over 30 providers and only one need this function. That does not sound like a good abstraction. 

> May be a solution is to just extend GeocodeQuery and send that. If I make a PR where would that extended GeocodeQuery live?

Since this is specific feature for MapQuest. I suggest you use `GeocodeQuery::withData` and make your changes only to the MapQuest provider.

[...]

PR #840, opened 2018-04-03
MapQuest improvement

TerjeBr:
This is an improvment to the MapQuest provider, that we talked about in #839 

I have tried to stay away from the code in the `Common` directory, but I could not avoid some small minor changes there to make it all fit together. Hope this will be ok.


Nyholm:
Thank you for this PR. 
I did a quick review and I see some issues when using a chain provider. 
Do you _really_ need a custom implementation of a query?

TerjeBr:
The custom implementation of a query is only to make it easier to use. It will work fine with a reqular query as well, as I wrote in the Readme.md

Nyholm:
I would be way happier without the custom implementation. That would also avoid the changes in Common

TerjeBr:
I am sorry that you do not like the custom implementation of `GeocodeQuery`. Is there something in particular that bothers you about it? Something I can improve?

Or are you just opposed to having two `GeocodeQuery` implementations in general?

The additional one should work fine as a replacement for all the cases where the original one may be used.

Without the new `GeocodeQuery` implementation, you first have to create a `GeocodeQuery` with a dummy address string (it cannot be empty), and the clone it in order to have the address data on it. I found this to be a bit cumbersome, so that is why I added the new implementation.

TerjeBr:
It is a bit unfortunate that the interface `Geocoder\Provider\Provider` (that all the providers must implement) depends on the concrete classes `Geocoder\Query\GeocodeQuery` and `Geocoder\Query\ReverseQuery` instead of interfaces. When in addition those concrete classes are `final` it leaves no room for changing or improving the Query classes that are used in a graceful way.

I guess changing it to use interfaces is the `right` thing to do, but that will be a huge BC break. Removing the `final` keyword on the concrete classes is less invasive. If you choose to do that, you also might as well make the properties `protected` instead of `private`.

The only other alternative I can see is to declare that the Query classes are set in stone, and we can have no suggested changes or development on the Query classes. I think that would be a bit unfortunate thing to do.

TerjeBr:
Hi @Nyholm what/when will be the next step?

src/Common/Query/GeocodeQuery.php
```diff
@@ -46,12 +46,12 @@
     /**
      * @var array
      */
-    private $data = [];
+    protected $data = [];
 
     /**
      * @param string $text
      */
-    private function __construct(string $text)
```
Nyholm: Why do you need to override a private constructor? Why not override the `create` function?

src/Provider/MapQuest/MapQuest.php
```diff
@@ -56,46 +84,91 @@
...
+            $params = $this->addOptionsForPostQuery($params, $options);
+
+            return $this->executePostQuery(static::GEOCODE_ENDPOINT, $params);
         }
+    }
 
-        return $this->executeQuery($url);
+    private function extractAddressFromQuery(CommonGeocodeQuery $query)
+    {
+        if ($query instanceof GetAddressInterface) {
```
Nyholm: Sorry for my lack of understanding. Why do we need this check? 

Dont we get the address either way? The way you implemented `getAddress` is: 

```php
public function getAddress()
{
      return $this->getData(static::DATA_KEY_ADDRESS);
}
````

File: src/Common/Query/GeocodeQuery.php [hunk 3]
```diff
@@ -46,12 +46,12 @@
     /**
      * @var array
      */
-    private $data = [];
```
Nyholm: Why do you need to modify the scope of `$data`? You could just introduce a new property on `Geocoder\Provider\MapQuest\GeocodeQuery`, right?

src/Common/Query/GeocodeQuery.php
```diff
@@ -19,14 +19,14 @@
...
  */
-final class GeocodeQuery implements Query
+class GeocodeQuery implements Query
 {
     /**
      * The address or text that should be geocoded.
      *
      * @var string
      */
-    private $text;
```
Nyholm: Is this needed? If you overwrite `withText` then there is no need for changing this to protected, right?

src/Common/Query/GeocodeQuery.php
```diff
@@ -19,14 +19,14 @@
 /**
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
-final class GeocodeQuery implements Query
```
Nyholm: This hurts a little but Im fine with it =)

src/Http/Provider/AbstractHttpProvider.php
```diff
@@ -60,6 +61,11 @@ protected function getUrlContents(string $url): string
         $request = $this->getMessageFactory()->createRequest('GET', $url);
         $response = $this->getHttpClient()->sendRequest($request);
 
+        return $this->parseHttpResponse($response, $url);
+    }
+
+    protected function parseHttpResponse(ResponseInterface $response, string $url): string
```
Nyholm: Thank you for this function.

src/Provider/MapQuest/GeocodeQuery.php
```diff
@@ -0,0 +1,58 @@
...
+            $address = $this->getAddress();
+            if ($address instanceof Location) {
+                $this->text = $this->formatAddress($address);
+            }
+        }
+
+        return $this->text;
+    }
+
+    protected function formatAddress(Location $address): string
```
Nyholm: This function seams to violate the single responsibility principle.

src/Provider/MapQuest/GeocodeQuery.php
```diff
@@ -0,0 +1,58 @@
...
+            parent::__construct($address);
+        }
+    }
+
+    public static function createFromAddress(Location $address): self
+    {
+        return new self($address);
+    }
+
+    public function getAddress()
```
Nyholm: This function is only use internally. It could be removed.

src/Provider/MapQuest/GeocodeQuery.php
```diff
@@ -0,0 +1,58 @@
...
+        if ($address instanceof Location) {
+            $this->data[static::DATA_KEY_ADDRESS] = $address;
+        } else {
+            parent::__construct($address);
+        }
+    }
+
+    public static function createFromAddress(Location $address): self
+    {
+        return new self($address);
```
Nyholm: Just do: 

```php
$foo = new self('');
return $foo->withData(static::DATA_KEY_ADDRESS], $address);
```

src/Common/Query/GeocodeQuery.php
TerjeBr: It seemed to me to be logical that when the class is not `final` any more, child classes should be free to do a different `__construct`. But I can do as you say, and override the create function instead, no problem.

src/Provider/MapQuest/MapQuest.php
TerjeBr: This was my attempt at future proofing the code a bit.
This way for any `GeocodeQuery` object that implements the `GetAddressInterface` we just call the `getAddess` method, and we do not care where it is stored. If I could have changed the common `GeocodeQuery` class I would have left it at that.

The line
```
return $query->getData(GeocodeQuery::DATA_KEY_ADDRESS);
```
is just a fallback for if you are using a regular `GeocodeQuery` that does not implement the `GetAddressInterface`.

I think it is better to depend on interfaces instead of depending on a known constant in a specific class.

I implemented `getAddress` the way I did to have max interoparability with the common `GeocodeQuery` class, so that a programmer did not *have* to use the mapquest version of the `GeocodeQuery` class.

src/Common/Query/GeocodeQuery.php
TerjeBr: My goal was that the new `GeocodeQuery` class could be a drop in and exchangeable with the old one. If I introduce a new property, I would also have to reimplement the withData method right?

It seemed easier for me to do it this way. If you agree to remove the `final` keyword, why cannot all properties be protected?

src/Common/Query/GeocodeQuery.php
TerjeBr: Yes, I can overwrite all the methods to avoid changing the properties in the base class to from private to protected. But I thought the proper way was to change the visibilty of the properties, not to duplicate the code in the methods?

src/Common/Query/GeocodeQuery.php
TerjeBr: Thank you for that. I am glad that you will at least allow this.

src/Http/Provider/AbstractHttpProvider.php
TerjeBr: You are welcome.

src/Provider/MapQuest/GeocodeQuery.php
TerjeBr: It is a bit more ineffecient, as it will create two objects instead of one, and then leave the first one to be garbage collected.
But yes, that is a valid way to do it if the methods in the base class cannot be changed to be protected instead of private.

src/Provider/MapQuest/GeocodeQuery.php
TerjeBr: How is that?

It is a protected method reuired by the above `getText()` method.
Should not a query containing an address have a method to convert that address to a oneline text query, so that the query can be used also in other providers?

TerjeBr:
Thank you for your review. I will wait for your response before I push a new set of changes.

src/Provider/MapQuest/GeocodeQuery.php
TerjeBr: This is only meant as a default value for the text of the query, and will not be called if the user calls `GoecodeQuery::withText` him/her self to set the text portion of the query.

(Btw. "seams" = sömmar,  "seems"= verkar)

TerjeBr:
@Nyholm any updates on this? I am still waiting for your comments.

TerjeBr:
@Nyholm Have you been on vacation or something?

Please continue the conversation so we can land this PR.

src/Provider/MapQuest/GeocodeQuery.php
Nyholm: It is really not that inefficient. We are talking optimization on less than duoble quotes vs single quotes.

TerjeBr: Ok, forget my inefficency arguments then. Please respond to the other points I raised.

TerjeBr:
@Nyholm if we are going to change the `Common\Query\GeocodeQuery` class from final to non-final, do you not agree that it makes sense to also change the constructor and all the properties from `private` to `protected`?

Do you agree with the above statement? Or do you have any reason I am not aware of for wanting to keep it all as private?

src/Common/Query/GeocodeQuery.php
Nyholm: I would like to keep the changes to the GeocodeQuery to a strict minimum.

TerjeBr:
Ok, If you will please accept #849 that I have made with the minimal changes to the common files, I will make the necessary changes to this PR to reflect that.

Hopefully we can then move this PR forward.

TerjeBr:
@Nyholm for me to make the tests pass you need to tag a new version of https://github.com/geocoder-php/provider-integration-tests

TerjeBr:
@nyholm plase make a new version tag on the [geocoder-php/provider-integration-tests](https://github.com/geocoder-php/provider-integration-tests) so that I can update the composer.lock files so this will do the tests properly.

You will also need to merge #849 and add a tag to that before this PR can show green on all tests.

TerjeBr:
@Nyholm, have you gone on vacation again?

willdurand:
I am locking this PR until we have a chance to review it. I am sorry but there is no need to ping/comment every day, that won't speed up the process.

Nyholm:
Im sorry Terje for not replying as quickly as you would like me to. 

When I said: 

> Try you restrict yourself to not update the common objects. You may send another PR with these changes later.

I meant: Please update this PR so you do not change common objects. 

I did **not** mean: Make a new PR with changes in common to be merged before this one. 

The common files has over 30 different providers. They all are happy with the current abstraction. Im sure that even the `MapQuest` could find a way to use them in the current state. 

For a future PR (after this one is merged), Im happy to review changes in common that you can prove would significantly improve maintainability and user experience for at least a handful of providers.

TerjeBr:
I have complied with everything you have requested, as far as I know.

If you want to merge just this pull request, that is fine, then you can ignore #849 
Or if you want do do it in stages, you can merge #849 first, and then this one.

Please tell me if there is anything else you need for this to be merged.

Nyholm:
> Please tell me if there is anything else you need for this to be merged.

I'll try to be extra clear with what I mean. 

This is the process I want to follow: 
1) You update this PR so you do not modify any classes in common
2) I review it and after I'm happy I'll merge it
3) (optional) You make a PR with changes in common that you can prove would significantly improve maintainability and user experience for at least a handful of providers.

TerjeBr:
Does that mean you are backtracking on what we already have discussed?

On 13th of April you agreed that I could at least change the `Common\Query\GeocodeQuery` by removing the `final` keyword. All the other changes I can code around, but I cannot code around that.

Less important are the changes to `Http\Provider\AbstractHttpProvider` (you even thanked me for adding a function there) and a small bugfix in `Common\Model\AddressBuilder`.
That is all the changes that you will find in #849 

So, have you changed your stance to not allow any changes at all in the common classes now?

TerjeBr:
And again we have this great wall of silence.

Is it too much to expect that you are able to respond to me within a day or so?
Does it have to be weeks between when we exchange comments on this PR? I find that very frustrating.

Nyholm:
> And again we have this great wall of silence.

Im doing open source work on my free time because I enjoy it. Doing open source is however not the *only* thing Im doing before and after normal work. 
Im not sure if I intrepid your messages correctly, but I sense a hostile tone. (Correct me if Im wrong.) That makes it no fun to spend time helping you get your PR merged. 

> So, have you changed your stance so to not allow any changes at all in the common classes now?

I've not changed my mind. But I feel that you do not really listen when I say "restrict yourself from making changes in common." I know it is possible to add the features you want to add without any changes at all in common (as explained). I want you to see that as well. Since I want to help you to get this merged, I now ask you to not make any changes at all in common since that will ease reviewing and the decisions to merge is not too critical for the rest of the project. 

You may think Im unfair or unreasonable. I do not want to discuss this anymore, but Im happy to review an updated PR.
If other maintainers think do not agree with me you are more welcome to say so.

TerjeBr:
> I know it is possible to add the features you want to add without any changes at all in common (as explained).

This is not true.

It is impossible for me to continue without at least removing the `final` keyword from the `Common\Query\GeocodeQuery` class. I thought we already discussed that, and that you agreed to that change.

All the other ones I can code around, but not that one.

TerjeBr:
Shall I go ahead and make the PR so that the removing of the `final` keyword is the only change among the common files?

TerjeBr:
You are aware that I updated this PR on 14th of May right? And that that update made it only contain 3 small changes to the common files that I thought we were in agreement on?

If you were not aware of that, that may be the cause that we are talking past each other?

TerjeBr:
@Nyholm please answer me so I can go ahead with this PR

TerjeBr:
Since you do not give any answers, I will just have to assume. I assume then that I reached the correct conclusion when I said "make the PR so that the removing of the final keyword is the only change among the common files". I will go ahead and do just that.

TerjeBr:
Ok now everything should be as you want it.

Please tag a new version for https://github.com/geocoder-php/provider-integration-tests
if you want all the tests to pass before you merge.

unkind:
> Without the new GeocodeQuery implementation, you first have to create a GeocodeQuery with a dummy address string (it cannot be empty), and then clone it in order to have the address data on it. I found this to be a bit cumbersome, so that is why I added the new implementation.

Do I understand right that you want to remove `final` just because you're missing named constructor like `GeocodeQuery::location(...)`?

unkind:
@Nyholm argh, removing `final` is a very bad sign in most cases. You can call me next time :)

unkind:
After reading original topic, I think you can just make (static) function to create `GeocodeQuery` for `MapQuest`: `MapQuest::createQuery(...)`.

By the way, why does Geocoder have `Query` interface? I thought you decided you throw off it: https://github.com/geocoder-php/Geocoder/pull/583/commits/40bdb8d9d78a52e88c95b5856b7ee9a216e476d1

Nyholm:
I've added a PR to your PR: https://github.com/TerjeBr/Geocoder/pull/1

TerjeBr:
Ok, I am back from my vacation. Let's see what has happened here.

TerjeBr:
@unkind wrote:
> Do I understand right that you want to remove final just because you're missing named constructor like GeocodeQuery::location(...)?
> 
> Maybe it's better to improve GeocodeQuery model instead of inheriting it?

You are right that it would be better to improve `GeocodeQuery` instead of inheriting it. That is what I asked about first in #839, if there could be a way to incorporate the address in the query. But that seems to be a no go from @Nyholm.

So, this quest of mine to inherit from `GeocodeQuery`  is my attempt to circumvent that so mapquest users could have at least some level of user friendliness in the use of this liberary. But now that hope seems to be thwarted too, since he has obviously changed his mind and wont even allow the `final` keyword to be removed from the class (that he previously agreed to).

TerjeBr:
Ok, I hope this can finally be merged now then.

It fails the style guide check, but that what changes that @Nyholm introduced, so I guess that is ok.

TerjeBr:
Ping

PR #845, opened 2018-04-23
add adminLevels to reverse geocoder

stevevance:
I was wondering why when using ArcGIS Online as a reverse geocoder would not return the United States state and the county. It turns out that the provider wasn't even returning an "adminLevels" array, which is what every other reverse geocode provider returns. I have added `adminLevels` to the returned object of a reverse geocode.


src/Provider/ArcGISOnline/ArcGISOnline.php
```diff
@@ -137,6 +137,13 @@ public function reverseQuery(ReverseQuery $query): Collection
         $county = !empty($data->Subregion) ? $data->Subregion : null;
         $countryCode = !empty($data->CountryCode) ? $data->CountryCode : null;
 
+        $adminLevels = [];
+        foreach (['Region', 'Subregion'] as $i => $property) {
```
Nyholm: Isn't it any more keys than these?

jbelien:
Hello @stevevance , any progress here ?

Nyholm:
Closing due inactivity.

PR #846, opened 2018-04-23
fix which Mapzen property becomes adminLevel 2

stevevance:
The `county` property from a Mapzen result should become `adminLevel` 2, not the `locality` property. This brings it in line with Bing and ArcGIS Online (the other reverse geocoders I tested).

### DIFF
--- a/src/Provider/Mapzen/Tests/MapzenTest.php
+++ b/src/Provider/Mapzen/Tests/MapzenTest.php
@@ -72,7 +72,7 @@ class MapzenTest extends BaseTestCase
         $this->assertEquals('Acklam Road', $result->getStreetName());
         $this->assertEquals('London', $result->getLocality());
         $this->assertCount(4, $result->getAdminLevels());
-        $this->assertEquals('London', $result->getAdminLevels()->get(2)->getName());
+        $this->assertEquals('London', $result->getAdminLevels()->get(3)->getName());
         $this->assertEquals('Kensington and Chelsea', $result->getAdminLevels()->get(1)->getName());
         $this->assertEquals('United Kingdom', $result->getCountry()->getName());
         $this->assertEquals('GBR', $result->getCountry()->getCode());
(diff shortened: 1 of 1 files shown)


stevevance:
I don't understand why the Travis CI build failed; I'm new to using this.

Nyholm:
Great. This PR looks fine. The original author of this provider did not know if the `county` key (or it did not exits back then). 

There are some test failing. Take the first failing as an example: 
```
1) Geocoder\Provider\Mapzen\Tests\MapzenTest::testReverseWithRealCoordinates
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'England'
+'Lancaster'
/home/travis/build/geocoder-php/Geocoder/src/Provider/Mapzen/Tests/MapzenTest.php:104
/home/travis/build/geocoder-php/Geocoder/vendor/phpunit/phpunit/phpunit:53
```

On [MapzenTest.php line 104](https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/Mapzen/Tests/MapzenTest.php#L104) we have the line 


```php
$this->assertEquals('England', $result->getAdminLevels()->get(3)->getName());
```
But since you modified the array from `$result->getAdminLevels()` you also need to modify this line. It should work if you did `->get(4)`

-------

The next error is on [MapzenTest.php:145](https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/Mapzen/Tests/MapzenTest.php#L145). It assumes a size of the `$result->getAdminLevels()`. You should just modify that assertion. 

You should do the same with the rest of the failures.

jbelien:
The Travis CI build fails because of those failed assertions :

- [`MapzenTest.php#L104`](https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/Mapzen/Tests/MapzenTest.php#L104) : "*Failed asserting that two strings are equal. (expected : 'England' - actual: 'Lancaster')*"
- [`MapzenTest.php#L145`](https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/Mapzen/Tests/MapzenTest.php#L145) : "*Failed asserting that actual size 3 matches expected size 2.*"
- [`MapzenTest.php#L192`](https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/Mapzen/Tests/MapzenTest.php#L192) : "*Failed asserting that actual size 4 matches expected size 3.*"

jbelien:
@stevevance I fixed the assertions. Travis CI build do not fail anymore.

PR #854, opened 2018-06-11
add new provider for Algolia Places

Fabex:
New Provider to use Algolia Places

## TODO
 - [x] Make Tests


src/Provider/AlgoliaPlaces/AlgoliaPlaces.php
```diff
@@ -0,0 +1,223 @@
...
+    const TYPE_BUS_STOP = 'busStop';
+
+    const TYPE_TRAIN_STATION = 'trainStation';
+
+    const TYPE_TOWN_HALL = 'townhall';
+
+    const TYPE_AIRPORT = 'airport';
+
+    /** @var string */
+    const ENDPOINT_URL = 'http://places-dsn.algolia.net/1/places/query';
```
Nyholm: Do we really want this? If you do not mind I suggest to remove it.

src/Provider/AlgoliaPlaces/AlgoliaPlaces.php
```diff
@@ -0,0 +1,223 @@
...
+            self::TYPE_COUNTRY,
+            self::TYPE_ADDRESS,
+            self::TYPE_BUS_STOP,
+            self::TYPE_TRAIN_STATION,
+            self::TYPE_TOWN_HALL,
+            self::TYPE_AIRPORT,
+        ];
+    }
+
+    protected function getUrlContents(string $url): string
```
Nyholm: Coudn't you use the method on AbstractHttpProvider?

src/Provider/AlgoliaPlaces/AlgoliaPlaces.php
Fabex: in fact in a lot of provider there is a constant like this one

GMaps
https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/GoogleMaps/GoogleMaps.php#L37

For OSM it's a little different but near the same way
https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/GoogleMaps/GoogleMaps.php#L37

TomTom
https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/GoogleMaps/GoogleMaps.php#L37

So I think we can keep it

Nyholm: I mean we should not use HTTP but only HTTPS. Im fine with you using a constant.

src/Provider/AlgoliaPlaces/AlgoliaPlaces.php
Fabex: In fact in the AbstractHttpProvider the request make a GET request, but for Algolia we need to make a POST request.

Nyholm: Check the latest changes (made yesterday) in `AbstractHttpProvider`. There is a new `getRequest` function that you want to override.

src/Provider/AlgoliaPlaces/AlgoliaPlaces.php
Fabex: oh, ok. I was not understood your question. So yes I'm totally agree with I will remove it

src/Provider/AlgoliaPlaces/AlgoliaPlaces.php
Fabex: Indeed, you have split the getUrlContents in 2 parts. So of course now I will just override getRequest and stop to duplicate some code 😄.

Nyholm: Yes. Correct.

Baachi:
@Nyholm Looks good to me, can i merge them?

Nyholm:
It does look good. But it still missing some tests, right?

jbelien:
There is indeed no test yet in this PR.
I think it's because of https://github.com/Fabex/geocoder-algolia-places-provider/pull/1

@Fabex @sheub Can you include your tests in this PR ? Thanks :)

Fabex:
Hi @Nyholm 
Tests are finally done. Big thanks to @sheub for his work.
I hope it's good for you

jbelien:
Awesome !
Thanks a lot guys !

I'll have a look ASAP 👍

Issue #853, opened 2018-06-07
Provide valid User-Agent for Nominatim


eileenmcnaughton:
Per https://github.com/eileenmcnaughton/org.wikimedia.geocoder/issues/1 it has been reported to us that 

"The nominatim.openstreetmap.org states that users must "provide a valid HTTP Referer or User-Agent identifying the application (stock User-Agents as set by http libraries will not do)."

The User-Agent is currently set to the default GuzzleHttp one. This is in violation of the usage policy."

Is it currently possible to control the user-agent with the Nominatim plugin to avoid this?

jbelien:
Thanks for the notice !

I'll have a look at this to se how we could fix the Nominatim provider to be compliant with the usage policy !

As an OpenStreetMap Belgium board member, I feel a bit ashamed about this... 🙈

eileenmcnaughton:
That's has made me smile. Also - I like your avatar! Everything is awesome

jbelien:
I just had a look at it !
I can fix it ; really not an issue but I need advice from the team ( ping @Nyholm @willdurand ) !

To set `Referer` and `User-Agent` headers, I'll need to replace the use of `getUrlContents()` from the parent `AbstractHttpProvider` class by its own (like I did for my provider here : https://github.com/geo6/geocoder-php-geo6-provider/blob/master/Geo6.php#L173-L198).

I was looking at the best way to do it and I found this PR #353 !
If we could add `getHeaders()`, `getHeader()`, and `setHeaders()` to the `AbstractHttpProvider` class, it would save us the trouble to "overload" the `getUrlContents()` function for each provider that needs to set headers.

What do you think about this ?
Thanks a lot !

eileenmcnaughton:
Looks good to me :-) But I guess you need @Nyholm @willdurand input

jbelien:
> Looks good to me :-) But I guess you need @Nyholm @willdurand input

Indeed :)
If I don't have any news, I'll just submit a PR on Nominatim provider without adding the "headers" functions to `AbstractHttpProvider` class in the next few days.

Nyholm:
Thank you for this issue. 

The "correct" way of doing this is something like: (not sure if "correct" is the correct word...)

```php
use Geocoder\Query\GeocodeQuery;
use Geocoder\Query\ReverseQuery;
use Http\Client\Common\Plugin\HeaderDefaultsPlugin;
use Http\Client\Common\PluginClient;

[...]

PR #859, opened 2018-06-21
Nominatim geocoder cannot handle IP addresses

mtmail:
Nominatim has no such feature.

The tests sent `127.0.0.1` and expected `localhost` in the response but that response is hardcoded in the AbstractProvider and never came from Nominatim.

Adjusted the tests. `phpunit` passes for me.


mtmail:
Makes sense. I added a changelog entry now and upgrade from 4.1=>5.0.

I assume PHP packagist takes the version number directly from the changelog? At least I couldn't find any other mention of 4.1 in other files.

Nyholm:
I (or some other maintainer) have to make a release. Packagist looks at all the releases. 

I think this looks good. I’ll to a more detailed review later. Or maybe @jbelien wants to take a look?

jbelien:
@Nyholm Let me just fix the issue with the headers (User-Agent and Referer) - see #853 - before creating a new release, I'll take care of that in the next few days 😃

jbelien:
PR #873 is ready to be reviewed.
Once both PR #859 and #873 are merged I think we can release version `5.0` of **Nominatim** provider !👍

jbelien:
I think we can merge this PR and release version `5.0` ! 🎉

PR #864, opened 2018-06-27
Split getUrlContents

Stadly:
As requested in https://github.com/geocoder-php/Geocoder/pull/829. Since `getUrlContents()` takes the url as a string argument, it is not possible to set headers, such as `Accept-Language`, for the request. Therefore I created `getRequest()` to generate the request based on the url string, but without fetching the contents. Then headers can be added to the request, before the request is passed to `getRequestContents()` which fetches the content.

### DIFF
--- a/src/Http/Provider/AbstractHttpProvider.php
+++ b/src/Http/Provider/AbstractHttpProvider.php
@@ -19,6 +19,7 @@ use Geocoder\Provider\AbstractProvider;
 use Http\Message\MessageFactory;
 use Http\Discovery\MessageFactoryDiscovery;
 use Http\Client\HttpClient;
+use Psr\Http\Message\RequestInterface;
 
 /**
  * @author William Durand <william.durand1@gmail.com>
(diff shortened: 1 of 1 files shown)


src/Http/Provider/AbstractHttpProvider.php
```diff
@@ -57,7 +58,32 @@ public function __construct(HttpClient $client, MessageFactory $factory = null)
      */
     protected function getUrlContents(string $url): string
     {
-        $request = $this->getMessageFactory()->createRequest('GET', $url);
+        $request = $this->getRequest($url);
+
+        return $this->getRequestContents($request);
```
Nyholm: Maybe rename this to parseResponse.

src/Http/Provider/AbstractHttpProvider.php
```diff
@@ -57,7 +58,32 @@ public function __construct(HttpClient $client, MessageFactory $factory = null)
...
+     *
+     * @param RequestInterface $request
+     *
+     * @return string
+     *
+     * @throws InvalidServerResponse
+     */
+    protected function getRequestContents(RequestInterface $request): string
+    {
         $response = $this->getHttpClient()->sendRequest($request);
```
Nyholm: I don’t think this function should send the request.

Stadly: We could change `getRequestContents(RequestInterface $request)` to `parseResponse(ResponseInterface $response)`, and call `$response = $this->getHttpClient()->sendRequest($request)` inside `getUrlContents` before calling `parseResponse`. The drawback would be that the request url would not be available inside `parseResponse`, so the method would require an extra parameter `$url` in order to throw `InvalidServerResponse`.

src/Http/Provider/AbstractHttpProvider.php
Stadly: See my reply to your other comment.

src/Http/Provider/AbstractHttpProvider.php
Nyholm: Oh. I see. Hm. You *could* throw the exception and the catch it...

If you decide to still send the request in this function, I still think you should rename it. Maybe to “getparsedResponse” or something similar. A request does not really have “content”. It has a body of it is a POST request.

Stadly: Good point. I'll rename it. If a provider needs to access the response, it would be possible to move `$response = $this->getHttpClient()->sendRequest($request);` into a separate method `sendRequest`, that could be overridden by the provider.

PR #872, opened 2018-06-30
changed 3 more tests

stevevance:


stevevance:
@Nyholm There is now only [one test](https://travis-ci.org/geocoder-php/Geocoder/jobs/398594014) that's failing...I changed the `adminLevel` that it's checking from 3 to 4 but it's still failing. 

Which leads me to question what API URL is being called. If it's Mapzen.com, it'll fail because that API doesn't work anymore. `Geocode.earth` is a drop-in replacement, but where do I change the URL to `geocode.earth`?

stevevance:
@jbelien I've updated the URL, and it still fails.

src/Provider/Mapzen/Tests/MapzenTest.php
```diff
@@ -99,9 +99,9 @@ public function testReverseWithRealCoordinates()
         $this->assertEquals('Gage Street', $result->getStreetName());
         $this->assertEquals('LA1 1UH', $result->getPostalCode());
         $this->assertEquals('Lancaster', $result->getLocality());
-        $this->assertCount(4, $result->getAdminLevels());
+        $this->assertNotEmpty($result->getAdminLevels());
         $this->assertEquals('Lancashire', $result->getAdminLevels()->get(1)->getName());
-        $this->assertEquals('England', $result->getAdminLevels()->get(3)->getName());
+        $this->assertEquals('England', $result->getAdminLevels()->get(4)->getName());
```
Nyholm: What if you change this back to 3?

Inspect and see what level that would give you "England".

stevevance: How do I run a test on my computer to see what levels are returned?

Nyholm: You can run only Mapzen by
```
cd src/Provider/Mapzen
composer update
./vendor/bin/phpunit
```

jbelien:
This PR is not necessary anymore, I fixed the `MapzenTest.php` file in #846 .

PR #873, opened 2018-07-02
Update Nominatim.php

### LINKED ISSUE #853: Provide valid User-Agent for Nominatim

jbelien:
Add `User-Agent` and `Referer` headers.
Fix #853

### DIFF
--- a/src/Provider/Nominatim/Nominatim.php
+++ b/src/Provider/Nominatim/Nominatim.php
@@ -55,7 +55,7 @@ final class Nominatim extends AbstractHttpProvider implements Provider
      */
     public static function withOpenStreetMapServer(HttpClient $client, string $userAgent, string $referer = '')
     {
-        return new self($client, 'https://vps.geocode.be/test.php', $userAgent, $referer);
+        return new self($client, 'https://nominatim.openstreetmap.org', $userAgent, $referer);
     }
 
     /**
--- a/src/Provider/Nominatim/Nominatim.php
+++ b/src/Provider/Nominatim/Nominatim.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 namespace Geocoder\Provider\Nominatim;
 
 use Geocoder\Collection;
+use Geocoder\Exception\InvalidArgument;
 use Geocoder\Exception\InvalidServerResponse;
 use Geocoder\Exception\UnsupportedOperation;
 use Geocoder\Location;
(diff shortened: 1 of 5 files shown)


jbelien:
Travis failed for PHP7 deps=low but it seems to be caused by GeoIP provider and not be related to Nominatim provider.

src/Provider/Nominatim/Nominatim.php
```diff
@@ -210,7 +210,16 @@ private function executeQuery(string $url, string $locale = null): string
...
         }
 
-        return $this->getUrlContents($url);
+        $request = $this->getRequest($url);
+
+        if (isset($_SERVER['HTTP_USER_AGENT'])) {
+            $request = $request->withHeader('User-Agent', $_SERVER['HTTP_USER_AGENT']);
+        }
+        if (isset($_SERVER['HTTP_REFERER'])) {
+            $request = $request->withHeader('Referer', $_SERVER['HTTP_REFERER']);
```
Nyholm: Is Referer need?

pnorman:
This PR *allows* the referer to be set, but doesn't appear to actually set it, so does it fix #853 by itself?

jbelien:
@Nyholm Yeah, I also asked myself the same question ...
And you're probable right, not a good thing to use global variables but the User-Agent or Referer **needs** to be set : 
> Provide a valid HTTP Referer or User-Agent identifying the application (stock User-Agents as set by http libraries will not do).

@pnorman @Nyholm What would you the best way to do it you think ?

src/Provider/Nominatim/Nominatim.php
jbelien: User-Agent **or** Referer needs to be set :
> Provide a valid HTTP Referer or User-Agent identifying the application (stock User-Agents as set by http libraries will not do).

Nyholm:
I think the best is to not use Referer and to let the user configure User-Agent with constructor injection.

jbelien:
Constructor (and tests) updated !
I also added the [Nominatim Usage Policy link](https://operations.osmfoundation.org/policies/nominatim/) to the README file.

---
> This PR allows the referer to be set, but doesn't appear to actually set it, so does it fix #853 by itself?

@pnorman I checked, the referer is actually set !

Nyholm:
Excellent. Thank you

PR #874, opened 2018-07-02
changed endpoint to a compatible third-party because Mapzen is gone

stevevance:


stevevance:
The error is "handshake failed". The URL works fine, perhaps it's because the `Geocode.earth` needs a key to use to test.

jbelien:
Please do not change Mapzen provider URL ; create a new "geocode.earth" provider instead.
Thanks.

Nyholm:
I agree with @jbelien. 

@stevevance, could you please copy to Mapzen provider to create a new Geocode.earth provider?

Issue #831, opened 2018-02-19
Create provider for Mapbox


chriscalip:

chriscalip:
https://github.com/mapbox/mapbox-gl-geocoder/issues/146

jbelien:
Actually it's more https://www.mapbox.com/api-documentation/#geocoding than the `mapbox-gl-geocoder` JavaScript library :)

I'll have a look when I have a bit of free time, it shouldn't be to difficult to add this provider :)

fedeisas:
I started working on this: https://github.com/fedeisas/mapbox-provider

I'm not sure I made the right choices when it comes to converting Mapbox features into Address components. But it's a start.

How can we make it an official provider @willdurand?

Nyholm:
When you are happy with your implementation, feel free to add a link to in in the readme. 

Im not sure we have any specific requirements for including it in this repo. But send a PR and I'll review it.

fedeisas:
@Nyholm done :-)

PR #876, opened 2018-07-09
Add Mapbox Provider

fedeisas:
Issue: #831


jbelien:
Great job, thanks a lot ! 👍 

It will be release as provider in the next few days.

PR #878, opened 2018-07-31
Allow to extend the ArcGISOnline class

pfrenssen:
I'm working for an organisation that is an ArcGIS Enterprise customer. They are not using the public endpoint (https://geocode.arcgis.com/) but have their own customized endpoint.

Currently I cannot use the ArcGISOnline provider since the endpoint is hardcoded and the class is final so it cannot be overridden.

This PR makes it possible to override the provider so that ArcGIS Enterprise customers can customize the functionality to match their individual needs.

### DIFF
--- a/src/Provider/ArcGISOnline/ArcGISOnline.php
+++ b/src/Provider/ArcGISOnline/ArcGISOnline.php
@@ -27,7 +27,7 @@ use Http\Client\HttpClient;
 /**
  * @author ALKOUM Dorian <baikunz@gmail.com>
  */
-final class ArcGISOnline extends AbstractHttpProvider implements Provider
+class ArcGISOnline extends AbstractHttpProvider implements Provider
 {
     /**
      * @var string
@@ -42,7 +42,7 @@ final class ArcGISOnline extends AbstractHttpProvider implements Provider
     /**
      * @var string
      */
-    private $sourceCountry;
+    protected $sourceCountry;
 
     /**
      * @param HttpClient $client        An HTTP adapter
@@ -166,7 +166,7 @@ final class ArcGISOnline extends AbstractHttpProvider implements Provider
      *
      * @return string
      */
-    private function buildQuery(string $query, int $limit): string
+    protected function buildQuery(string $query, int $limit): string
     {
         if (null !== $this->sourceCountry) {
             $query = sprintf('%s&sourceCountry=%s', $query, $this->sourceCountry);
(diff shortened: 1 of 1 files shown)


jbelien:
Hello @pfrenssen ,

Thanks a lot for your PR !

I must admit I don't know how a local instance of ArcGIS Enterprise work but is the URL to the geocoding service similar to the official one (or is it completely different) ?

Official one : `https://geocode.arcgis.com/arcgis/rest/services/World/GeocodeServer/find?text=%s`
Local one (?) : `https://myownserver/arcgis/rest/services/World/GeocodeServer/find?text=%s`

pfrenssen:
For the moment it looks to be the same, but I am not 100% sure. Our documentation is just an introduction that explains basic usage and is referring to the original ArcGISOnline documentation in case we need to know more specific details. It doesn't mention anything about deviating from the API.

jbelien:
If it's the same, I think it will be better (and easier) to do like we did for Nominatim provider (and other) : adding the abilitiy to specify the url in the constructor and set a "default" url to the "official" one.

You can see how we did it here : [Nominatim.php#L57](https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/Nominatim/Nominatim.php#L57)

But @Nyholm and @willdurand may have other insight about this 😄

pfrenssen:
Having a configurable endpoint like is done for the Nominatim provider would solve my problem.

I am wondering though, what is the reasoning behind making these providers `final`? If I take a quick look at the current implementation I can see several reasons why a developer might want to override it:

* The implementation is not complete. For example the parameters `featureTypes`, `locationType` etc are not implemented. A developer might want to override the constructor and `::buildQuery()` methods if they need to use one of these optional parameters.
* Some options are hardcoded. By making the class final these cannot be overridden. For example the data format is hardcoded to `json` in `::buildQuery()`. A developer might prefer to set this to `pjson` to get a pretty printed version which is suitable for generating test data.

Nyholm:
It is final because it is an implementation of an interface. 

> The implementation is not complete.

=) I read this as "Do not put `final` on classes that are not perfect". 

> Some options are hardcoded.

That is no good. I would appreciate PRs that help the ArcGISOnline to be better and more flexible.

pfrenssen:
You are right, the correct solution would not be to make the class extendable, but to implement the missing parameters, and make sure that all options are configurable rather than hard coded.

I will close this since the suggested solution is wrong. We can solve these individual problems in dedicated tickets.

Thanks for the insights!

Issue #860, opened 2018-06-24
freegeoip.net will be discontinued on July 1, 2018


mwisner:
According to https://github.com/apilayer/freegeoip#readme the freegeoip.net service will be discontinued on July 1st, 2018. 

It's suggested in that announcement to use ipstack which doesn't appear to be supported by this package. 

Should the freegeoip provider be removed and / or ipstack added?

Nyholm:
Thank you. 
I suggest to add ipstack. There is no need to remove the provider at this time.

jGielen:
https://github.com/geocoder-php/Geocoder/pull/879

jGielen:
@mwisner in the mean time you can change the url

jbelien:
The ipstack provider is now merged and will be release in the next few days 🎉

PR #879, opened 2018-08-11
adding provider ipstack

jGielen:
adding ipstack as a provider, replacing freegeoip that is deprecated


jGielen:
I added ipstack for issue #860 but I can't get it to pass, don't see why.

jbelien:
Great job, thanks a lot ! 👍 

It will be release as provider in the next few days.

PR #880, opened 2018-08-16
Improve error message in case a provider can NOT geocode an address / query

holtkamp:
in case a provider can NOT geocode an address / query

### DIFF
--- a/src/Provider/Chain/Chain.php
+++ b/src/Provider/Chain/Chain.php
@@ -55,8 +55,12 @@ final class Chain implements Provider, LoggerAwareInterface
             } catch (\Throwable $e) {
                 $this->log(
                     'alert',
-                    sprintf('Provider "%s" could geocode address: "%s".', $provider->getName(), $query->getText()),
-                    ['exception' => $e]
+                    'Provider "{providerName}" could not geocode address: "{address}".',
+                    [
+                        'exception' => $e,
+                        'providerName' => $provider->getName(),
+                        'address' => $query->getText(),
+                    ]
                 );
             }
         }


holtkamp:
@Nyholm to be honest, I did not check. This one appeared in my logs and seemed "strange" to me 😉

holtkamp:
@Nyholm when thinking of it, I personally would use:

```php 
$logger->alert('Some informational message', $context = [ ... ]);
```
instead of 
```php 
$logger->log('alert', 'Some informational message', $context = [ ... ]);
```
I suppose LoggerInterface::log() should only be used when the severity level is a variable / determined dynamically...

Nyholm:
You may be correct. I do not mind either way.

holtkamp:
Will have a look to all logging messages coming time and see where such improvements can be applied in a separate PR. This is more a "bug fix" in my opinion, as the message was plain wrong 🤓

Nyholm:
Okey. Thank you

holtkamp:
@Nyholm any chance this PR lands in a "patch" version? I guess `4.0.1` (has it been that long since a release? Impressive stability!

holtkamp:
@Nyholm is a new patch release something which can be considered?

jbelien:
Sorry @holtkamp , I think we forgot about this.
I'll release `geocoder-php/chain-provider` `4.0.1` ASAP ! 👍

Nyholm:
Yeah, that sounds like a great idea. 

Could we do a small PR to the change log of the chain provider?

jbelien:
> Could we do a small PR to the change log of the chain provider?

Already on it :P

jbelien:
Done : https://packagist.org/packages/geocoder-php/chain-provider#4.0.1

Thanks for the reminder @holtkamp 🎉

holtkamp:
mm, I now see that the reverse query logging has the same "issue": https://github.com/geocoder-php/chain-provider/blob/master/Chain.php#L106-L109

jbelien:
Indeed, I'm on it ! :)

jbelien:
@holtkamp Done: https://packagist.org/packages/geocoder-php/chain-provider 🎉

simonschaufi:
@holtkamp  Hi, I have the problem, that the placeholders don't get replaced any more in my error message after this change.
My error message looks like this: `Provider "{providerName}" could not geocode address: "{address}".`. I don't see any point where the placeholders get replaced. I guess something like sprintf is missing (which was there before but got dropped).

holtkamp:
Hi @simonschaufi, what implementation of the PSR Logger are you using?

When reading the specs it seems that support for "placeholders" is an optional feature. 

https://www.php-fig.org/psr/psr-3/

I use https://github.com/Seldaek/monolog which *does* support placeholders...

simonschaufi:
Hi @holtkamp, I use Laravel Framework which has a wrapper Class for `monolog/monolog` (installed: 1.24.0). I debugged the whole logging and don't find the point where there is any replacement within the message

holtkamp:
And does your Monolog logger use the [PsrLogMessageProcessor](https://github.com/Seldaek/monolog/blob/master/doc/02-handlers-formatters-processors.md#processors)

The documentation states:

>Processes a log record's message according to PSR-3 rules, replacing {foo} with the value from $context['foo'].

simonschaufi:
This is the sub package: https://github.com/illuminate/log Couldn't find it there. What exactly is missing to make that work? I will make a PR then to add the feature in Laravel

holtkamp:
Well, I am not very familiar with Laravel, but some searches give me:

https://stackoverflow.com/questions/51694036/laravel-5-6-how-to-push-monolog-processor
https://hotexamples.com/examples/illuminate.log/Writer/getMonolog/php-writer-getmonolog-method-examples.html


But maybe we need to revert to using `sprintf()` instead of `{placeholders}`. I was not aware that support for placeholders is optional for PSR Loggers. @jbelien what do you think?

PR #896, opened 2018-10-28
Google Maps component filtering

Farskies:
Hi!

First of all, thanks for a great library.

I noticed that [component filtering](https://developers.google.com/maps/documentation/geocoding/intro#ComponentFiltering) was missing for the Google Maps provider, so i took the liberty of implementing it. Not sure if this can cause any issues with backwards compatibility.

### DIFF
--- a/src/Provider/GoogleMaps/GoogleMaps.php
+++ b/src/Provider/GoogleMaps/GoogleMaps.php
@@ -115,6 +115,11 @@ final class GoogleMaps extends AbstractHttpProvider implements Provider
             $url .= sprintf('&bounds=%s,%s|%s,%s', $bounds->getSouth(), $bounds->getWest(), $bounds->getNorth(), $bounds->getEast());
         }
 
+        if (null !== $components = $query->getData('components')) {
+            $serializedComponents = is_string($components) ? $components : $this->serializeComponents($components);
+            $url .= sprintf('&components=%s', urlencode($serializedComponents));
+        }
+
         return $this->fetchUrl($url, $query->getLocale(), $query->getLimit(), $query->getData('region', $this->region));
     }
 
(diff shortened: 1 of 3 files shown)


jbelien:
Hello @Farskies ,

Thanks a lot for your contribution ! 👍 

I'll have a look at your PR ASAP ! It looks very interesting !

Could you first make your PR pass the Travis build ? Thanks.

Farskies:
Hi!

I've rebased master to fix the failing tests. I've also added additional tests to check that the actual uri format is correct.

I'm not sure about the test using the real API, since it might not be reliable.

Tell me if you want anything changed.

.gitignore
```diff
@@ -4,3 +4,4 @@ composer.phar
 phpunit.xml
 php-cs-fixer.phar
 .puli/
+.idea/
```
jbelien: Is this really necessary ?

Farskies: It's just a habit from using PhpStorm, which automatically creates the `.idea` folder in the project root. I can remove it if you wish.

jbelien: I would prefer to update only the relevant files :)

Farskies:
You're right. If you want the library to either throw an exception on invalid keys or just plain out filtering them out, that's definitely possible.

Another possibility would be letting the API handle the invalid keys. That way, the library doesn't need to be updated when keys are added and/or changed. I'm unsure whether the API silently filters invalid keys or returns an error response, but I'll investigate it.

Which do you think is better?

jbelien:
If the API handles correctly (silent filter) I guess that's okay to let the Geocoder Provider send everything without any check.
But if the API returns an error if there is an unexpected component, I think it's better to filter it before sending the request to the API.

@Nyholm @willdurand what's your opinion on this ? :)

jbelien:
@Farskies Did you have time to check how the API handles invalid keys ?

Thanks !

Farskies:
I've been super busy lately, but I'll have a look at it today and update the pull-request accordingly!

Farskies:
Hi @jbelien!

Sorry for the big delay. I've confirmed that the Google Maps API silently filters out any invalid component names.

```
curl -G --data-urlencode "address=high st hasting" --data-urlencode "components=country:GB|invalid:invalid" --data-urlencode "key=x" https://maps.googleapis.com/maps/api/geocode/json
```

Returns the same data with and without the invalid component name.

I've also removed `.idea` from the gitignore, and rebased my branch.

PR #900, opened 2018-11-08
Improve Nominatim provider

jbelien:
- Migration from format XML to JSON (v2)
- Add `countrycodes` geocoding query parameter
- Add `viewbox` (and `bounded`) geocoding query parameter(s)

See https://wiki.openstreetmap.org/wiki/Nominatim

I paid attention to not create any breaking change.


src/Provider/Nominatim/Model/NominatimAddress.php
```diff
@@ -73,24 +73,34 @@ public function withAttribution(string $attribution = null): self
     /**
      * @return null|string
      */
+    public function getCategory()
+    {
+        return $this->category;
+    }
+
     public function getClass()
```
Nyholm: We should mark this function as `@deprecated`

src/Provider/Nominatim/Model/NominatimAddress.php
```diff
@@ -73,24 +73,34 @@ public function withAttribution(string $attribution = null): self
...
+    public function withCategory(string $category = null): self
     {
         $new = clone $this;
-        $new->class = $class;
+        $new->category = $category;
 
         return $new;
     }
 
+    public function withClass(string $category = null): self
```
Nyholm: Why do we need this? Dont we rename `class` to `category`?

src/Provider/Nominatim/Nominatim.php
```diff
@@ -54,7 +54,7 @@
      *
      * @return Nominatim
      */
-    public static function withOpenStreetMapServer(HttpClient $client, string $userAgent, string $referer = '')
+    public static function withOpenStreetMapServer(HttpClient $client, string $userAgent, string $referer = ''): self
```
Nyholm: I think adding this is a BC break. Or is this class `final`?

jbelien: I think this class could be declared as `final` !
But I can remove the `self`, it doesn't "add" a lot.

src/Provider/Nominatim/Model/NominatimAddress.php
jbelien: Done !

src/Provider/Nominatim/Nominatim.php
jbelien: I just checked : the class is (and was) indeed `final` : https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/Nominatim/Nominatim.php#L33

jbelien:
@Nyholm Can I merge this ? :)

src/Provider/Nominatim/Model/NominatimAddress.php
Nyholm: Oh, Yes of course. 

The diff was weird. I thought you added a new "withClass" function

jbelien:
Thanks ! 👍

PR #905, opened 2018-11-23
Create Photon provider

jbelien:
See https://photon.komoot.de/


src/Provider/Photon/Model/PhotonAddress.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        return $this->osmId;
+    }
+
+    /**
+     * @param null|int $osmId
+     *
+     * @return PhotonAddress
+     */
+    public function withOSMId(int $osmId = null): self
```
eugenekurasov: What are you think about to use this structure `withOSMId(?int $osmId)`.

With current structure I can call with no pass argument `withOSMId()`.

src/Provider/Photon/Photon.php
```diff
@@ -0,0 +1,187 @@
...
+     * @var string
+     */
+    private $rootUrl;
+
+    /**
+     * @param HttpClient $client an HTTP client
+     *
+     * @return Photon
+     */
+    public static function withKomootServer(HttpClient $client)
```
eugenekurasov: Maybe add default value for  $rootUrl in `__construct`?
 ` __construct(HttpClient $client, $rootUrl = 'https://photon.komoot.de')`

src/Provider/Photon/Model/PhotonAddress.php
jbelien: That would be better indeed but Geocoder is suppose to be compatible with PHP 7.0 (see [here](https://github.com/geocoder-php/Geocoder/blob/master/composer.json#L15)) and [Nullable types](http://php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types) is a PHP 7.1 feature.

src/Provider/Photon/Photon.php
jbelien: That could be an option indeed !
I don't really have an opinion on this but I like the fact that by using `Photon::withKomootServer()` the developer knows for a fact that he will use server from Komoot (instead of its own for instance).

eugenekurasov: Why you don't add return type declaration for this method `withKomootServer(HttpClient $client): self` ?

jbelien: That's indeed missing 👍

src/Provider/Photon/composer.json
```diff
@@ -0,0 +1,42 @@
...
+    "keywords": [],
+    "homepage": "http://geocoder-php.org/Geocoder/",
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Jonathan Beliën"
+        }
+    ],
+    "require": {
+        "php": "^7.0",
```
Nyholm: Why not 7.2?
That is the lowest active version

jbelien: I changed to PHP 7.2 but then I realized that Geocoder itself requires PHP 7.0 (see https://github.com/geocoder-php/Geocoder/blob/master/composer.json#L15).

Wouldn't it make more sense to keep PHP 7.0 as requirement for providers too ?

jbelien: Confirm by Travis : build is made with PHP 7.0 and 7.1 :)

jbelien:
@Nyholm Can we release this ?
I can take care of it but you'll need to create a new "sub-module" if I recall correctly.

PR #906, opened 2018-11-26
[Mapbox] Fix Mapbox Query builder when there is Bound

damienalexandre:
When using Bound, the query was malformed in two ways:

- the format was wrong (`|` is not allowed);
- ~~the order was wrong (`minLon,minLat,maxLon,maxLat` was not respected).~~

I also added some documentation to the "Bound" object because it may not be clear to everyone what the min latitude can be when reading compass points.

### DIFF
--- a/src/Common/Model/Bounds.php
+++ b/src/Common/Model/Bounds.php
@@ -40,10 +40,10 @@ final class Bounds
     private $east;
 
     /**
-     * @param float $south
-     * @param float $west
-     * @param float $north
-     * @param float $east
+     * @param float $south South bound, also min latitude
+     * @param float $west  West bound, also min longitude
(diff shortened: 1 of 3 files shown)


jbelien:
Thanks @damienalexandre ! 👍

damienalexandre:
Oh gosh, you are right. I used [this image](https://gis.stackexchange.com/questions/255158/get-minimum-and-maximum-latitude-and-longitude-of-new-york) as reference to build my test query but it's plain wrong... 

PR Updated, commits squashed.

jbelien:
[New release](https://github.com/geocoder-php/mapbox-provider/releases/) of Mapbox provider 🎉

damienalexandre:
Awesome thank you for your reactivity! :clap:

PR #922, opened 2018-12-22
Run backwards compatibillity check

Nyholm:
We can test each provider on each PR if they break BC.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,8 @@ matrix:
         - php: 7.0
         - php: 7.1
         - php: 7.2
+        - php: 7.3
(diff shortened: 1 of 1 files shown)


jbelien:
Oh nice, that's awesome 🎉

Nyholm:
Hm. 

Im not sure this is working.
The monoloth repo has no version to benchmark from. And the provider packages they dont have a `.git` folder in this repo. 

We *could* add this on the tests on each provider. But that would obviously not check any PRs only show as a warning before we tag new versions.

Nyholm:
:(

PR #924, opened 2018-12-26
Update GoogleMapsTest.php with API key

### LINKED ISSUE #923: Google Maps: Keyless access to Google Maps Platform is deprecated.
> You must have a valid API key and a billing account to access our APIs. When you enable billing, you will get $200 free usage every month for Maps, Routes, or Places. Based on the millions of users using our APIs today, most of them can continue to use Google Maps Platform for free with this credit. Having a billing account helps us understand our developers’ needs better and allows you to scale seamlessly.

Since 11 June 2018, we can't query the Google Maps Geocoding API without a key.

That means that tests that are not cached will fail with the following "error".
```json
{
   "error_message" : "Keyless access to Google Maps Platform is deprecated. Please use an API key with all your API calls to avoid service interruption. For further details please refer to http://g.co/dev/maps-no-account",
   "results" : [],
   "status" : "OVER_QUERY_LIMIT"
}
```

I noticed this while working on reviewing PR #918 ; I noticed that Travis was failing and (at first) didn't understood why.

I think it's worth updating the tests by adding API key every time (I'll submit a PR in a few minutes) and update the cached responses (**but I don't have a Google Maps API key** so I can't do it).

@Nyholm @willdurand (or even @Tiraniel) : would you have a Google Maps API key I could use the update the tests (or you could even do it yourself) ? Thanks.

jbelien:
Add Google Maps API key for every (not mocked) tests because "Keyless access to Google Maps Platform is deprecated."

Close #923

### DIFF
--- a/src/Provider/GoogleMaps/Tests/GoogleMapsTest.php
+++ b/src/Provider/GoogleMaps/Tests/GoogleMapsTest.php
@@ -70,7 +70,11 @@ class GoogleMapsTest extends BaseTestCase
      */
     public function testGeocodeWithRealIp()
     {
-        $provider = new GoogleMaps($this->getHttpClient());
+        if (!isset($_SERVER['GOOGLE_GEOCODING_KEY'])) {
+            $this->markTestSkipped('You need to configure the GOOGLE_GEOCODING_KEY value in phpunit.xml');
+        }
+
+        $provider = new GoogleMaps($this->getHttpClient($_SERVER['GOOGLE_GEOCODING_KEY']), null, $_SERVER['GOOGLE_GEOCODING_KEY']);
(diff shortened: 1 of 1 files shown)


jbelien:
🚨 Before merging it: `src\Provider\GoogleMaps\Tests\.cached_responses` needs to be cleaned and tests need to be updated !

**Requires a Google Maps API key**

Nyholm:
Tests are updated with new responses. We are ready to merge =)

jbelien:
Awesome, thanks a lot @Nyholm 👍 

Unable to merge though because of the failing Photon test (due to PHP 7.0 vs PHP 7.1).

PR #927, opened 2019-01-05
[Photon] Require PHP 7.1

Nyholm:
PHP 7.0 is not longer supported. 

This PR will also trigger a push to https://github.com/geocoder-php/photon-provider

### DIFF
--- a/src/Provider/Photon/composer.json
+++ b/src/Provider/Photon/composer.json
@@ -11,7 +11,7 @@
         }
     ],
     "require": {
-        "php": "^7.0",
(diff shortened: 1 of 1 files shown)


jbelien:
Seems good !
But we should switch to PHP 7.1 for all providers and main components then IMHO :)

Nyholm:
Some people consider that a breaking change. 
Doctrine does not tough..

Im okey with updating PHP versions only when we already plan a new major release. Like now for the Photon provider.

PR #928, opened 2019-01-05
Remove nyholm/psr7 from require-dev

Nyholm:


.travis.yml
```diff
@@ -18,9 +18,9 @@ env:
 matrix:
     fast_finish: true
     include:
-        - php: 7.0
+        - php: 7.2
           env: deps="low"
-        - php: 7.1
+        - php: 7.2
           env: deps="high"
         - php: 7.0
```
jbelien: Shouldn't we remove that `php: 7.0` line ?

Nyholm: No, We still want to test on 7.0

Issue #793, opened 2017-11-03
Administrative level 4 is defined twice


benwrigley:
Hi There,

Using the Google API and ReverseQuery with coordinates sometimes  dies  with Geocoder\Exception\InvalidArgument 'Administrative level 4 is defined twice'

Google docs do say this might happen. Is it possible to merge the two lists for the purposes of returning data and therefore not having to generate an Exception?

Nyholm:
Hey. 

Thank you for this issue. 
You are correct. What do you suggest we should do if we notice multiple level 4’s?

benwrigley:
Hi,

I wonder if there is a way to merge the lists in the level 4s?

benwrigley:
so take this example:

`        {
          "long_name": "Kings Hill",
          "short_name": "Kings Hill",
          "types": [
            "administrative_area_level_4",
            "locality",
            "political"
          ]
        },
        {
          "long_name": "Kings Hill",
          "short_name": "Kings Hill",
          "types": [
            "administrative_area_level_4",
            "political"
          ]
        },`

The types would all be merged and uniqued so the result might be ['locality','political']

Nyholm:
Hm.. A bunch of questions comes to my mind.
- Should we always merge?
- Should we only merge if they are exactly the same?
- Should we really correct Google? If Google does not care about this issue, should we?

Nyholm:
Closed because of inactivity.

geodeveloper:
Same, here **Administrative level 3 is defined twice**. I guess the correct solution would be to get the first administrative area level of the same type.

fritzmg:
Same here, with __Administrative level 1 is defined twice__ with the following query:
```
{“text”:“\u017delezni\u010dn\u00e1 53 Stupava 90031",“bounds”:“null”,“locale”:null,“limit”:5,“data”:[]}
```
and following result:
```
{
   "results":[
      {
         "address_components":[
            {
               "long_name":"53",
               "short_name":"53",
               "types":[
                  "street_number"
               ]
            },
            {
               "long_name":"\u017delezni\u010dn\u00e1",
               "short_name":"\u017delezni\u010dn\u00e1",
               "types":[
                  "route"
               ]
            },
            {
               "long_name":"Stupava",
               "short_name":"Stupava",
               "types":[
                  "political",
                  "sublocality",
                  "sublocality_level_1"
               ]
            },
            {
               "long_name":"Stupava",
               "short_name":"Stupava",
               "types":[
[...]

PR #939, opened 2019-02-28
[Google Maps] do not use duplicate sub locality levels

fritzmg:
Fix for #793

If an `AdminLevel` has the exact same code, name and level, it will lead to an 
```
Administrative level x is defined twice
``` 
exception in willdurand\geocoder. This PR filters such `AdminLevel`s. It also works with
```php
$subLocalityLevels = array_filter($subLocalityLevels, function($adminLevel) use ($subLocalityLevels) {
    foreach ($subLocalityLevels as $localityLevel) {
        if ($adminLevel == $localityLevel) {
            return false;
        }
    }
    return true;
});
```
of course, but I am not sure what the best practise for object comparison is.

### DIFF
--- a/src/Provider/GoogleMaps/Model/GoogleAddress.php
+++ b/src/Provider/GoogleMaps/Model/GoogleAddress.php
@@ -499,11 +499,11 @@ final class GoogleAddress extends Address
             $subLocalityLevels[] = new AdminLevel($level['level'], $name, $level['code'] ?? null);
         }
 
-        $subLocalityLevels = array_filter($subLocalityLevels, function($adminLevel) use ($subLocalityLevels) {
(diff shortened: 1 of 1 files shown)


jbelien:
Thanks @fritzmg !

Before reviewing your PR, can you make sure it passes every test ?
Thanks.

fritzmg:
Sure, no problem, I'll check the failling test.

jbelien:
Sorry, I just had a look at your PR and I can not accept it because it breaks the `testReverseWithSubLocalityLevels()` test !

**Detail:**
Admin levels should be:
- Level 1 = Iijima
- Level 4 = 58

Your PR returns **no** admin level.

jbelien:
I submitted a new PR (#943) that should fix issue #793

fritzmg:
Ah thank you - I unfortunately never got around to look at this.

PR #943, opened 2019-04-12
[Google Maps] Fix "Administrative level X is defined twice"

jbelien:
This should fix the "Administrative level X is defined twice" error (see #793).

@benwrigley @geodeveloper @fritzmg Could one of you test it and confirm ? Thanks


geodeveloper:
Thanks for the PR. Will try to test it but as far as I can see in the file changes it should work just fine.

PR #950, opened 2019-04-30
Allow to configure log levels

holtkamp:


src/Provider/Chain/Chain.php
```diff
@@ -35,9 +45,21 @@ final class Chain implements Provider, LoggerAwareInterface
     /**
      * @param Provider[] $providers
      */
-    public function __construct(array $providers = [])
+    public function __construct(array $providers = [], string $geocodeQueryLogLevel = 'alert', string $reverseQueryLogLevel = 'alert')
```
atymic: Do we need a different log level on both methods, or would be it be adequate to set it for the class?
Since we can't pass the logger to the constructor, i'd skip passing the log levels there.

src/Provider/Chain/Chain.php
```diff
@@ -27,6 +27,16 @@ final class Chain implements Provider, LoggerAwareInterface
...
 
+    /**
+     * @var string
+     */
+    private $geocodeQueryLogLevel;
+
+    /**
+     * @var string
+     */
+    private $reverseQueryLogLevel;
```
atymic:
```suggestion
    private $logLevel = LogLevel::ALERT;
```

src/Provider/Chain/Chain.php
holtkamp: >Do we need a different log level on both methods, or would be it be adequate to set it for the class?

I would "be happy" with the ability to differentiate between the two types of functionality:

- geocode
- reverse geocode

Other users also might have other use cases / needs. 

>Since we can't pass the logger to the constructor, i'd skip passing the log levels there.

Not sure how this relates with each other? I do agree we should use the `LogLevels::ALERT` constant as default. 

But feel

atymic: >I would "be happy" with the ability to differentiate between the two types of functionality:

On second thought, I guess there's no reason to reduce configurability.

> Not sure how this relates with each other? I do agree we should use the LogLevel::ALERT constant as default.

I was meaning, keeping it consistent by having the loggers set by setters, rather than passed in the constructor.

Keep in mind i'm not the owner of this repo, so take my ideas with a grain of salt :smile:

holtkamp: >I was meaning, keeping it consistent by having the loggers set by setters, rather than passed in the constructor.

Aah, ok, understood. Yeah, might be better to not pollute the constructor too much.

atymic:
cc @jbelien

holtkamp:
ping @atymic and @jbelien, can this be considered for merge?

holtkamp:
Thanks @jbelien 

Does this justify a new minor release? Our logs are flooded with alert log entries when using Geocoders that do not support certain addresses (for example IP-addresses). These are known limitations and no alerts need be triggered 🤓

jbelien:
Yes, I will release a new version for the Chain provider! 🎉 

But first I need to fix a (non related) issue : https://github.com/geocoder-php/Geocoder/issues/983 (to be able to merge my PR about the CHANGELOG) 😕

jbelien:
@holtkamp Done : https://packagist.org/packages/geocoder-php/chain-provider 🍾

Issue #951, opened 2019-04-30
What is the domain are of this project?


gnumoksha:
Hi. I'm a bit confused about the domain area of this project. Is this related to geocoding only?

Quoting [google maps docs](https://developers.google.com/maps/documentation/geocoding/start):

> Geocoding is the process of converting addresses (like a street address) into geographic coordinates (like latitude and longitude), which you can use to place markers on a map, or position the map.
> 
> Reverse geocoding is the process of converting geographic coordinates into a human-readable address.

Just to be clear, I need to enrich address data received from many different resources and I've found geocoding providers are designed for unambiguous address. Do all providers have this same rule?

jbelien:
From the README:
> Geocoder is a PHP library which helps you build geo-aware applications by providing a powerful abstraction layer for geocoding manipulations.

This library provides an easy way to query plenty of geocoding providers. Each geocoding provider has its own rules !
We provide a structured way to query every provider the same way and the result is also always structured the same way no matter what geocoding provider you chose.

Geocoder library includes geocoding, reverse geocoding but also "IP geocoding".

gnumoksha:
> This library provides an easy way to query plenty of geocoding providers. Each geocoding provider has its own rules !

I've quoted Google's definition of Geocoding but in fact, this is the global definition of Geocoding (see [Wikipedia](https://en.wikipedia.org/wiki/Geocoding))

> We provide a structured way to query every provider the same way and the result is also always structured the same way no matter what geocoding provider you chose.
> 
> Geocoder library includes geocoding, reverse geocoding but also "IP geocoding".

Right. What I'm saying is to clarify what this result would be. Besides Google Maps, the [Here](https://github.com/geocoder-php/here-provider) provider is also using geocoding and the [nominatim](https://github.com/geocoder-php/nominatim-provider) provider uses address + geocoding.

I'm saying all this because for many people the concept of "geocoding" is unclear. The fact of one need a nonambiguous address to search an address using geocoding is an issue that the README can clarify too. For example, I came to this issue with the geocoding definition because of my team at work had thought it was ok to search ambiguous address using geocoding, when in fact, it resulted in poor results.results.

jbelien:
> Right. What I'm saying is to clarify what this result would be. Besides Google Maps, the Here provider is also using geocoding and the nominatim provider uses address + geocoding.

I don't get this.

[...]

PR #952, opened 2019-04-30
chore(README): explain what is geocoding

gnumoksha:
See #951


atymic:
ping @jbelien 	

Maybe worth merging or closing this? 😄

jbelien:
I'm not sure it's useful/necessary to explain what geocoding is since this library does not geocode (but provides an abstraction layer).

And I definitely don't like just copy-pasting Google's definition.
Would much more prefer to copy-paste Wikipedia definition and a link to the Wikipedia page.

@Nyholm What do you think about this ?

atymic:
Closing this due to inactivity.

PR #957, opened 2019-05-16
Add new GraphHopper Provider

jbelien:


jbelien:
@Nyholm Could you create a new package for the GraphHopper provider ? Thanks :)

Nothing urgent though 😉

PR #964, opened 2019-06-01
chore: update geoip2 documentation

atymic:
This PR updates the docs for the geoip2 provider to make then clearer on how to use the database and web service.

I did consider adding a test for the API client, but unfortunately I don't think is possible due to the lack of dependency injection on the `Client` class.

cc @jbelien


atymic:
> Could you just revert the `CHANGELOG` ?
> No need for a new release, it's documentation update.
> Thanks.

Done :)

PR #972, opened 2019-06-24
fix: deprecate geoips provider

### LINKED ISSUE #965: Deprecating Shutdown Providers
Since the provider has been shut down for over a year, should we consider deprecating it?
http://freegeoip.net/shutdown

Maybe it would be a good idea to come up with a policy for handling provider shutdowns.

Possible steps:
- Mark packagist package as abandoned
- adding `@deprecated` annotations to the class & constructor (which will warn people using IDEs)
- Remove provider from main README
- Update provider README to indicate it's deprecation
- Skip/Remove tests (maybe)?

atymic:
- adds `@deprecated` annotations to the class & constructor (which will warn people using IDEs)
- Updates the readme and changelog to communicate the deprecation
- Removes it from the main readme.

Todo: 
- Mark packagist as abandoned.

Closes #965


atymic:
I've also added some documentation to the repo for how to deprecate providers. Let me know if you have any thoughts/suggestions.

src/Provider/GeoIPs/GeoIPs.php
```diff
@@ -27,10 +27,8 @@
 use Http\Client\HttpClient;
 
 /**
- * @author Andrea Cristaudo <andrea.cristaudo@gmail.com>
```
Nyholm: No need to remove these author tags

atymic: No worries, have added them back.

atymic:
No worries 😄

atymic:
@jbelien sorry to bother, but is there anything else that needs doing on this?

atymic:
(if you are approving again so I can merge, It doesn't give me the option unfortunately)

Issue #881, opened 2018-08-24
Chain Provider catch Exception


geodeveloper:
Hi,

Currently, the chain provider only logs the exceptions, there is no way to catch the exceptions and we always get an empty `AddressCollection` in case something went wrong. There is no way to know if the collection is empty because there were no results or if an error occurred.

A simple customized solution would be to extend the class `Chain` and throw the exception but it's not possible since it's `final`. All I can do is to replicate the class (not best solution).

If there is a workaround about this I would like to hear about it, otherwise I guess it must be implemented a way to catch the erros.

Thank you.

holtkamp:
Also encountered this: currently it is not possible to influence how the ChainProvider deals with exceptions. 

It would be nice to make it configurable to:
- rethrow caught Exceptions
- indicate the severity level of the logged exceptions, which is currently fixed at the `alert` level

*Rationale*
In a chain of Geocoders, one of them is not capable of geocoding IP-addresses. Ok, no problem, the next Geocoder in the chain might be able to do this. But this does result in a lot of `alert` level exceptions. It would be nice to either decrease the level of logging, and/or rethrow the Exception so an `UnsupportedOperation` could be caught and dealt with gracefully.

holtkamp:
Submitted a PR for the ability to configure the log level: https://github.com/geocoder-php/Geocoder/pull/950

chriscalip:
@geodeveloper Curious from quote "All I can do is to replicate the class (not best solution)." What is not the best solution nor even a workable for your app?

geodeveloper:
@chriscalip it's not the best solution because you need to replicate the whole class when maybe you only need to override one method. Now on future releases I have to maintain this custom class myself if anything changes on all other methods.

chriscalip:
@geodeveloper I see how about an alternative.. given that the interfaces wont be changing this major version; you can have a custom class with a composition of using the ChainProvider. On the custom class implementation methods call-in the ChainProvider .. oh wait logger needs to be the ChainProvider. Ok I got you.

atymic:
What about if we catch exceptions as we currently do, and then rethrow a `ChainException` which contains all of the exceptions encountered.

You'd then be able to loop over the caught exceptions and handle them as required.

atymic:
PR open in #978 with a potential solution.

PR #978, opened 2019-07-01
feat: save chain exceptions

atymic:
Relates to #881 

Would be interested in your thoughts @geodeveloper @Nyholm @jbelien 
This allows the chain to execute as it does currently, but adds the ability to handle the exceptions if they occured.

For example, you could log `QuotaExceeded` exceptions and ignore `UnsupportedOperation` ones.

Example:
```php
$chain->geocodeQuery($query);

if ($chain->getPreviousQueryExceptions()) {
    $exceptionsToLog = array_filter($chain->getPreviousQueryExceptions(), function (\Throwable $e) {
         return $e instanceof QuotaExceeded;
    })
}
```

### DIFF
--- a/src/Provider/Chain/Chain.php
+++ b/src/Provider/Chain/Chain.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 namespace Geocoder\Provider\Chain;
 
 use Geocoder\Collection;
+use Geocoder\Exception\LogicException;
 use Geocoder\Model\AddressCollection;
 use Geocoder\Query\GeocodeQuery;
 use Geocoder\Query\ReverseQuery;
@@ -33,6 +34,11 @@ final class Chain implements Provider, LoggerAwareInterface
     private $providers = [];
 
     /**
+     * @var \Throwable[]|null
+     */
+    private $previousQueryExceptions;
+
+    /**
      * @param Provider[] $providers
      */
     public function __construct(array $providers = [])
@@ -45,6 +51,8 @@ final class Chain implements Provider, LoggerAwareInterface
      */
     public function geocodeQuery(GeocodeQuery $query): Collection
     {
+        $this->previousQueryExceptions = [];
+
         foreach ($this->providers as $provider) {
             try {
                 $result = $provider->geocodeQuery($query);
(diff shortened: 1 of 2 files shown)


geodeveloper:
Hi @atymic,

Thanks for bringing up this solution. I guess that will work as a workaround. I think the class Chain needs a better approach to handle different scenarios and the `final` class should be reconsidered if it's really necessary.

Nyholm:
What if we collect exceptions and then if no result we throw either:
1) one exception if `count($exceptions) === 1`
2) A `new ChainExceptionCollection($exceptions)`

We do someting similar in Symfony/messanger.

atymic:
@Nyholm 

I'm not sure that would work in this context. The issue is the chain provider only runs until we get a valid result (which it returns), so we'd never have a chance to throw the exceptions.

For example:
```
Provider 1 -> Exception
Provider 2 -> Exception
Provider 3 -> Success

Results in return of $result from the last provider
```

Unless we changed the behaviour to throw an exception even if we find a valid result, the only case where we would throw a `ChainExceptionCollection` would be when they all fail.

Nyholm:
We should not throw exception if we get a result. They should only be thrown if no result was found. 

I like the idea of getting exceptions throw a getter even if we do find a result.

atymic:
> We should not throw exception if we get a result. They should only be thrown if no result was found.

I agree. I can add a `ChainExceptionCollection` in the case that no results are found, but that would be a BC break hence why I went with this solution (which doesn't break BC)

Nyholm:
No it wouldn’t. The contract is from the interface. The interface states that an exception might be thrown, right?

Finesse:
> I can add a ChainExceptionCollection in the case that no results are found

I want to clarify. A provider interface says that a provider returns an empty collection when there is no result for the query and no error has happened (e.g. when the given address doesn't exist). Therefore empty result doesn't mean that an exception must be thrown. I think that the chain provider must throw an exception when no results are retrieved and the inner providers haven't thrown an exception.

alexsegura:
> I think that the chain provider must throw an exception only when no results are retrieved and an inner provider has thrown an exception.

I agree with @Finesse. 

Well, actually, it depends on what a "chain" provider is intended to be used for. 
It is here to have a "winner takes all" behavior, or is it here to have a "fault tolerant" behavior?

Issue #942, opened 2019-04-10
[Feature Request] Lookup by id (GoogleMapsProvider, Nominatim)


franzwilding:
First of all, thanks for this great php package!

I have a feature request for getting address information by id instead of doing a full search again. I think many applications will store part of the geocoding response and eventually will re-fetch information for the stored part. 

At the moment, with GoogleMapsProvider and NominatimProvider you can only do a reverse lookup by lat/lng or re-do the query. However both of the endpoints have a dedicated API for lookup (for google maps this even is the preferred way, because it will not count as a request as far as I know). 

For OpenStreetMap it looks like this: https://nominatim.openstreetmap.org/lookup?osm_ids=NXXXXX.

It would be great if the providers would have a lookup method for this.
An even better solution would be to extend the provider Interface (or have a 2nd interface "LookupAwareProvider" with a lookup action or the GeocodeQuery object with an id property and maybe also the Location interface.

If a provider defines an id for a location, it could be used for a lookup call.

I know, that I can easily achieve this by creating my own provider, however, if this feature is interesting for you, we could discus an solution and I could provide a PR for this.

jbelien:
Hello @franzwilding ,

That's indeed (IMHO) a good idea !
I think we could add it to our [`Geocoder` Interface](https://github.com/geocoder-php/Geocoder/blob/master/src/Common/Geocoder.php) rather easily.

What do you think about this @willdurand  @Nyholm ?

---

**Documentation:**

- Nominatim: https://wiki.openstreetmap.org/wiki/Nominatim#Address_lookup
- Google Maps: https://developers.google.com/maps/documentation/geocoding/intro#place-id

franzwilding:
Hey @jbelien, 

I think an even better solution would be to extend the Gecoder Interface with an id-aware-geocoder interface, because some geocoders will not support an id, I guess?

jbelien:
You're right but that's not an issue.

Some providers do not support reverse geocoding either (have a look at [AlgoliaPlaces](https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/AlgoliaPlaces/AlgoliaPlaces.php#L100) for instance).
For consistency, I would add a `lookup` function to the standard `Geocoder` interace.

franzwilding:
Ok cool. I will try to find time to create a PR for this!

franzwilding:
[...]

PR #979, opened 2019-07-02
Feat #942 implement provider lookup method

franzwilding:


src/Common/Geocoder.php
```diff
@@ -53,4 +53,15 @@ public function geocode(string $value): Collection;
...
+    /**
+     * Lookup a location by it's provider-specific id.
+     *
+     * @param mixed $id
+     *
+     * @return Collection
+     *
+     * @throws \Geocoder\Exception\Exception
+     */
+    public function lookup($id): Collection;
```
atymic:
```suggestion
    public function lookup(string $id): Collection;
```

I think it would be good to type hint this as `string`. I

atymic: (and add this to all other places in the code)

src/Common/Query/LookupQuery.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        return ['id' => $this->id];
+    }
+
+    /**
+     * @return string
+     */
+    public function __toString()
+    {
+        return $this->id;
```
atymic: To be consistent with the other `Query` types I think this should return something like:

```php
    public function __toString()
    {
        return sprintf('LookupQuery: %s', json_encode([
            'id' => $this->getId(),
            'locale' => $this->getLocale(),
            'data' => $this->getAllData(),
        ]));
    }
```

src/Provider/GoogleMaps/Tests/GoogleMapsTest.php
```diff
@@ -175,6 +176,39 @@ public function testReverseWithRealCoordinates()
...
+     */
+    public function testLookup()
+    {
+        $provider = new GoogleMaps($this->getMockedHttpClient());
+        $provider->lookupQuery(new LookupQuery('1'));
+    }
+
+    public function testLookupWithRealId()
+    {
+        $place_id = 'Ei5TdmFydnN0b2xzdsOkZ2VuIDEwLCAxMjYgMzggSMOkZ2Vyc3RlbiwgU3dlZGVuIhoSGAoUChIJMV6ykqx3X0YRj43VlDVv6T0QCg';
```
atymic: Maybe worth camel casing here to comply with PSR2

atymic:
If you need any help getting this done let me know @franzwilding :)

jbelien:
We indeed need to update all the providers and add an `UnsupportedOperation` on the `lookupQuery` for those who do not support this operation (just like [`reverseQuery` in `AlgoliaPlaces `](https://github.com/geocoder-php/algolia-places-provider/blob/master/AlgoliaPlaces.php#L99), for instance)

src/Common/Geocoder.php
Nyholm: Adding this method here will break BC :/

Nyholm:
I dont think we can add a `UnsupportedOperation` . That would violate Liskov substitution principle. 

I think we should create a new interface for this new feature (instead of modifying `Geocoder`). Then maybe a new trait (instead of modifying GeocoderTrait). 

That will make it easy to add to providers one by one. (Or just the one we like). It will also avoid breaking BC.

mrelliot69:
was this ever pushed?

jbelien:
> was this ever pushed?

No, we are still waiting for an update from the author.

PR #986, opened 2019-07-23
[Chain] Improve error message when provider can not *reverse* geocode

jbelien:
Apply PR #880 for **reverse** geocoding


simonschaufi:
@jbelien Hi, I have the problem, that the placeholders don't get replaced any more in my error message after this change.
My error message looks like this: `'Provider "{providerName}" could not reverse geocode coordinates: {latitude}, {longitude}".',`. I don't see any point where the placeholders get replaced. I guess something like sprintf is missing (which was there before but got dropped).

PR #988, opened 2019-07-24
Yandex support key

### LINKED ISSUE #987: Yandex Provider ApiKey
Hello. Yandex geocoder api requires api key now. Can you add this feature?
[from](https://yandex.ru/blog/mapsapi/novye-pravila-dostupa-k-api-kart?from=tech_pp)

atymic:
This PR adds the ability to use an API with the `Yandex` provider, as per the API changes. See the issue: #987 

I also fixed all the tests that were failing.

Closes #987


src/Provider/Yandex/Yandex.php
```diff
@@ -60,14 +67,15 @@ public function __construct(HttpClient $client, string $toponym = null)
...
+
         // This API doesn't handle IPs
         if (filter_var($address, FILTER_VALIDATE_IP)) {
             throw new UnsupportedOperation('The Yandex provider does not support IP addresses, only street addresses.');
         }
 
         $url = sprintf(self::GEOCODE_ENDPOINT_URL, urlencode($address));
 
-        return $this->executeQuery($url, $query->getLocale(), $query->getLimit());
+        return $this->executeQuery($url, $query->getLimit(), $query->getLocale());
```
jbelien: Why switch the order ?

src/Provider/Yandex/Yandex.php
```diff
@@ -84,7 +92,7 @@ public function reverseQuery(ReverseQuery $query): Collection
             $url = sprintf('%s&kind=%s', $url, $toponym);
         }
 
-        return $this->executeQuery($url, $query->getLocale(), $query->getLimit());
+        return $this->executeQuery($url, $query->getLimit(), $query->getLocale());
```
jbelien: Why switch the order ?

src/Provider/Yandex/Yandex.php
```diff
@@ -96,18 +104,22 @@ public function getName(): string
...
-     * @param string|null $locale
-     * @param int         $limit
+     * @param string $url
+     * @param int    $limit
+     * @param string $locale
      *
      * @return AddressCollection
      */
-    private function executeQuery(string $url, string $locale = null, int $limit): AddressCollection
+    private function executeQuery(string $url, int $limit, string $locale = null): AddressCollection
```
jbelien: Why switch the order ?

atymic: As per PSR2:

> Method arguments with default values MUST go at the end of the argument list.

https://www.php-fig.org/psr/psr-2/#44-method-arguments

It's only a change on the private API, so I don't think it's a problem

jbelien: Sorry, I indeed didn't see that one of the 2 had a default value.

PR #991, opened 2019-07-29
[Here] Fix reverse geocoding (a typo was present in the prox parameter)

DamienHarper:
Fix reverse geocoding (a typo was present in the prox parameter)
Bump API version from 8 to 9


DamienHarper:
@jbelien I have no idea why 4 out of the 6 jobs are failing on Travis :/

atymic:
@DamienHarper they are failing because the URLs have changed, so not there are no cached responses for these requests.

You need to copy `phpunit.xml.dist` to `phpunit.xml` and add your HERE api keys. Then, delete all the cached responses are run the tests (which will generate/save the new cached responses).

Then you should commit those cache files, and travis will use them for the tests.

It would also be good to add a test to cover the bug you fixed, if possible.

DamienHarper:
@atymic thanks for the guidelines, everything's ok now

src/Provider/Here/Tests/HereTest.php
```diff
@@ -88,19 +101,19 @@ public function testReverseWithRealCoordinates()
         $results = $provider->reverseQuery(ReverseQuery::fromCoordinates(48.8632156, 2.3887722));
 
         $this->assertInstanceOf('Geocoder\Model\AddressCollection', $results);
-        $this->assertCount(1, $results);
+        $this->assertCount(5, $results);
```
jbelien: Any idea why this returns 5 results (instead of 1 previously) ?

DamienHarper: @jbelien because the URL was malformed: radius value in the URL was not appended to the prox parameter using a coma but using an ampersand so it was considered as a new parameter so it messed up the API call and it returned always one result that way (the maxresults parameter was even ignored)

jbelien:
Does the upgrade to version 9 of the API has an impact ?

DamienHarper:
No, they only added timezone information in the responses (cf https://developer.here.com/documentation/geocoder/topics/backwards-compatibility.html#backwards-compatibility__api-different-generations)

jbelien:
Thanks a lot @DamienHarper ! 🎉 

New release `0.2.1` is now available: https://packagist.org/packages/geocoder-php/here-provider

PR #995, opened 2019-08-14
[Nominatim] Fix enotices

eileenmcnaughton:
Current code expects osm_id & osm_type to be defined. However when I ran it with just a postcode they were not defined & I got an enotice - ie

this url
https://nominatim.openstreetmap.org/search/search?format=jsonv2&q=90210%2CUnited+States&addressdetails=1&limit=5

Gave me this

```
[
  {
    "place_id": 200404707,
    "licence": "Data © OpenStreetMap contributors, ODbL 1.0. https://osm.org/copyright",
    "boundingbox": [
      "34.085763939939",
      "34.085863939939",
      "-118.2905153602",
      "-118.2904153602"
    ],
    "lat": "34.0858139399386",
    "lon": "-118.290465360203",
    "display_name": "Los Angeles, California, 90210, United States of America",
    "place_rank": 21,
    "category": "place",
    "type": "postcode",
    "importance": 0.535,
    "address": {
      "city": "Los Angeles",
      "state": "California",
      "postcode": "90210",
      "country": "United States of America",
      "country_code": "us"
    }
  }
]
```

### DIFF
--- a/src/Provider/Nominatim/Nominatim.php
+++ b/src/Provider/Nominatim/Nominatim.php
@@ -230,8 +230,8 @@ final class Nominatim extends AbstractHttpProvider implements Provider
 
         $location = $builder->build(NominatimAddress::class);
         $location = $location->withAttribution($place->licence);
-        $location = $location->withOSMId(intval($place->osm_id));
-        $location = $location->withOSMType($place->osm_type);
+        $location = $location->withOSMId(intval($place->osm_id ?? NULL));
+        $location = $location->withOSMType($place->osm_type ?? NULL);
         $location = $location->withDisplayName($place->display_name);
 
         if (false === $reverse) {
--- a/src/Provider/Nominatim/Nominatim.php
+++ b/src/Provider/Nominatim/Nominatim.php
@@ -230,10 +230,10 @@ final class Nominatim extends AbstractHttpProvider implements Provider
 
         $location = $builder->build(NominatimAddress::class);
         $location = $location->withAttribution($place->licence);
-        if (!empty($place->osm_id) {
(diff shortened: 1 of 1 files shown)


src/Provider/Nominatim/Nominatim.php
```diff
@@ -230,8 +230,8 @@ private function jsonResultToLocation(\stdClass $place, bool $reverse): Location
 
         $location = $builder->build(NominatimAddress::class);
         $location = $location->withAttribution($place->licence);
-        $location = $location->withOSMId(intval($place->osm_id));
-        $location = $location->withOSMType($place->osm_type);
+        $location = $location->withOSMId(intval($place->osm_id ?? NULL));
```
atymic:
```suggestion
        $location = $location->withOSMId(intval($place->osm_id ?? null));
```
Lowercase `null` please as per PSR2 (and below)

atymic: I don't think the null fallback should be inside int cast.  Maybe something like:

```suggestion
if (!empty($place->osm_id) {
    $location = $location->withOSMId((int) $place->osm_id);
}
```

jbelien: `intval($place->osm_id ?? null)` is wrong anyway, it would set the `OSMId` to `0` which is not what we want!

eileenmcnaughton:
Hi I'm probably not up for writing a test - this is pretty tangental to what I'm working on & I addressed it in our code just by altering our test data (since it was a test failing in a use case that wasn't important to us). I just did the PR through the github UI

jbelien:
Superseded by #997.

PR #1002, opened 2019-08-15
fix: update freegeoip post-shutdown

atymic:
freegeoip.net shut down > 1 year ago, so I think it makes sense to update the provider.

Since it will still work with the open source, self hosted variant i've just updated the constructor default to `null` and throw an exception when trying to geocode without explicit setting of a host.

cc @jbelien

### DIFF
--- a/src/Provider/FreeGeoIp/Readme.md
+++ b/src/Provider/FreeGeoIp/Readme.md
@@ -10,7 +10,7 @@
 This is the Free GeoIp provider from the PHP Geocoder. This is a **READ ONLY** repository. See the
 [main repo](https://github.com/geocoder-php/Geocoder) for information and documentation. 
 
-# Freegeoip Shutdown
+## Freegeoip Shutdown
 As per the [freegeoip.net](http://freegeoip.net/shutdown) website, the provider has been purchased by [IpStack](https://ipstack.com/).
 As a result, this provider no longer works with the default configuration. It will still work if you use the 
 [self hosted variant](https://github.com/apilayer/freegeoip/) and supply a host when constructing the provider.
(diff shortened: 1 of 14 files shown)


src/Provider/FreeGeoIp/Readme.md
```diff
@@ -10,6 +10,24 @@
 This is the Free GeoIp provider from the PHP Geocoder. This is a **READ ONLY** repository. See the
 [main repo](https://github.com/geocoder-php/Geocoder) for information and documentation. 
 
+# Freegeoip Shutdown
+As per the [freegeoip.net](http://freegeoip.net/shutdown) website, the provider has been purchased by [IpStack](https://ipstack.com/).
+As a result, this provider no longer works with the default configuration. It will still work if you use the 
+[self hosted variant](https://github.com/apilayer/freegeoip/) and supply a host when constructing the provider.
+
```
jbelien: I think we should also add a link to https://github.com/geocoder-php/ipstack-provider :)

atymic: Updated :)

src/Provider/FreeGeoIp/Readme.md
```diff
@@ -10,6 +10,27 @@
 This is the Free GeoIp provider from the PHP Geocoder. This is a **READ ONLY** repository. See the
 [main repo](https://github.com/geocoder-php/Geocoder) for information and documentation. 
 
+# Freegeoip Shutdown
```
jbelien: I would change the titles "level": `##` (+1 level) here

src/Provider/FreeGeoIp/Readme.md
```diff
@@ -10,6 +10,27 @@
...
+$provider = new Geocoder\Provider\FreeGeoIp\FreeGeoIp($httpClient, 'http://my.internal.geocoder/json/%s');
+```
+
+## Alternatives
+We offer an [IpStack provider](https://github.com/geocoder-php/ipstack-provider) which you can use if you wish to continue with the new service owner.
+
+### Full IP Provider List
+https://github.com/geocoder-php/Geocoder#ip
+
 ### Install
```
jbelien: `##` (-1 level) here

jbelien: `##` (-1 level) for `### Contribute` title

src/Provider/FreeGeoIp/Readme.md
atymic: Updated (and the others)!

PR #1004, opened 2019-08-17
New Pelias provider

jbelien:
Pelias is an open-source geocoder. You can find its [documentation here](https://github.com/pelias/documentation).
Pelias does not provide an API per se but other API (and providers) are based on Pelias and extend this provider.

Geocode Earth will be updated to extend this new provider (see #1005).

### DIFF
--- a/src/Provider/Pelias/Pelias.php
+++ b/src/Provider/Pelias/Pelias.php
@@ -1,0 +1,242 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Geocoder package.
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @license    MIT License
+ */
+
+namespace Geocoder\Provider\Pelias;
+
+use Geocoder\Collection;
+use Geocoder\Exception\InvalidCredentials;
+use Geocoder\Exception\QuotaExceeded;
+use Geocoder\Exception\UnsupportedOperation;
+use Geocoder\Model\Address;
+use Geocoder\Model\AddressCollection;
+use Geocoder\Query\GeocodeQuery;
+use Geocoder\Query\ReverseQuery;
+use Geocoder\Http\Provider\AbstractHttpProvider;
+use Geocoder\Provider\Provider;
+use Http\Client\HttpClient;
+
+abstract class Pelias extends AbstractHttpProvider implements Provider
+{
+    /**
+     * @var string
+     */
+    private $root;
+
+    /**
+     * @var int
+     */
+    private $version;
+
+    /**
+     * @param HttpClient $client  an HTTP adapter
+     * @param string     $root    url of Pelias API
+     * @param int        $version version of Pelias API
+     */
+    public function __construct(HttpClient $client, string $root, int $version = 1)
+    {
+        $this->root = sprintf('%s/v%d', rtrim($root, '/'), $version);
+        $this->version = $version;
+
+        parent::__construct($client);
+    }
+
+    /**
+     * @param GeocodeQuery $query
+     * @param array        $query_data Additional query data (API key for instance).
+     *
+     * @return string
+     *
+     * @throws \Geocoder\Exception\Exception
+     */
+    protected function getGeocodeQueryUrl(GeocodeQuery $query, array $query_data = []): string
+    {
+        $address = $query->getText();
+
+        // This API doesn't handle IPs
+        if (filter_var($address, FILTER_VALIDATE_IP)) {
+            throw new UnsupportedOperation('The GeocodeEarth provider does not support IP addresses, only street addresses.');
+        }
+
+        $data = [
+            'text' => $address,
+            'size' => $query->getLimit(),
+        ];
+
+        return sprintf('%s/search?%s', $this->root, http_build_query(array_merge($data, $query_data)));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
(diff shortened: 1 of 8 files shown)


jbelien:
Should I leave it `abstract` or should I give the possibility to provide an URL here ?
Second option make it easier to use it (no need to create a new provider that extends this one).

What do you think @Nyholm @atymic ?

src/Provider/Pelias/Pelias.php
```diff
@@ -0,0 +1,251 @@
...
+     *
+     * @throws \Geocoder\Exception\Exception
+     */
+    protected function getGeocodeQueryUrl(GeocodeQuery $query, array $query_data = []): string
+    {
+        $address = $query->getText();
+
+        // This API doesn't handle IPs
+        if (filter_var($address, FILTER_VALIDATE_IP)) {
+            throw new UnsupportedOperation('The Pelias provider does not support IP addresses, only street addresses.');
```
atymic: Minor, but not sure if you want to get the provider name of the current provider here (vs the base provider).

src/Provider/Pelias/Pelias.php
```diff
@@ -0,0 +1,251 @@
...
+            switch ($json['meta']['status_code']) {
+                case 401:
+                case 403:
+                    throw new InvalidCredentials('Invalid or missing api key.');
+                case 429:
+                    throw new QuotaExceeded('Valid request but quota exceeded.');
+            }
+        }
+
+        if (!isset($json['type']) || 'FeatureCollection' !== $json['type'] || !isset($json['features']) || 0 === count($json['features'])) {
```
atymic: Maybe worth splitting this into a few lines to make it easier to read.

src/Provider/Pelias/Pelias.php
```diff
@@ -0,0 +1,251 @@
...
+        $locations = $json['features'];
+
+        if (empty($locations)) {
+            return new AddressCollection([]);
+        }
+
+        $results = [];
+        foreach ($locations as $location) {
+            $bounds = [
+                'south' => null,
```
atymic: Should we be initialising this to null?

> Changed: Location::getBounds will return null or a Bounds object with coordinates data. It will never return Bounds without data.

src/Provider/Pelias/Tests/PeliasTest.php
```diff
@@ -0,0 +1,143 @@
...
+     * @expectedExceptionMessage The Pelias provider does not support IP addresses, only street addresses.
+     */
+    public function testGeocodeWithLocalhostIPv6()
+    {
+        $provider = new Pelias($this->getMockedHttpClient(), 'http://localhost/');
+        $provider->geocodeQuery(GeocodeQuery::create('::1'));
+    }
+
+    /**
+     * @expectedException \Geocoder\Exception\UnsupportedOperation
```
atymic: I think `expectedException` is remove in the latest version of phpunit.
Maybe worth using `$this->expectedException()` instead?

src/Provider/Pelias/Pelias.php
```diff
@@ -0,0 +1,251 @@
...
+use Geocoder\Exception\UnsupportedOperation;
+use Geocoder\Model\Address;
+use Geocoder\Model\AddressCollection;
+use Geocoder\Query\GeocodeQuery;
+use Geocoder\Query\ReverseQuery;
+use Geocoder\Http\Provider\AbstractHttpProvider;
+use Geocoder\Provider\Provider;
+use Http\Client\HttpClient;
+
+class Pelias extends AbstractHttpProvider implements Provider
```
atymic: Is this the only provider that can be extended (all of the ones i've seen have `final`)

src/Provider/Pelias/Pelias.php
jbelien: Good idea 👍

src/Provider/Pelias/Pelias.php
jbelien: Nice catch, thanks!

src/Provider/Pelias/Pelias.php
jbelien: AFAIK, it's indeed the first (and only) provider that can be extended.

src/Provider/Pelias/Tests/PeliasTest.php
jbelien: That's a copy-paste from the other providers.
But you, indeed, seem to be right!

jbelien:
@atymic Do you think it's ready to be merged ?

src/Provider/Pelias/README.md
```diff
@@ -0,0 +1,33 @@
...
+[![Software License](https://img.shields.io/badge/license-MIT-brightgreen.svg?style=flat-square)](LICENSE)
+
+This is the Pelias provider from the PHP Geocoder. This is a **READ ONLY** repository. See the
+[main repo](https://github.com/geocoder-php/Geocoder) for information and documentation.
+
+## Pelias
+
+Pelias is an open-source geocoder. You can find its [documentation here](https://github.com/pelias/documentation).
+
+Pelias does not provide an API per se but other API (and providers) are based on Pelias and extend this provider.
```
atymic: Maybe worth noting that you can run your own server?

atymic:
> @atymic Do you think it's ready to be merged ?

I'm happy for it to be merged, looks good to me 😄

jbelien:
@Nyholm Could you create the sub-package for the new Pelias provider ? Thanks :)

PR #1005, opened 2019-08-17
Extend Geocode Earth provider from Pelias provider

jbelien:
Related to #1004 (waiting for it to be merged first)


src/Provider/GeocodeEarth/GeocodeEarth.php
```diff
@@ -52,22 +38,17 @@ public function __construct(HttpClient $client, string $apiKey)
         }
 
         $this->apiKey = $apiKey;
-        parent::__construct($client);
+        parent::__construct($client, 'https://api.geocode.earth/', 1);
```
atymic: Think it would be good to have these in constants still :)

jbelien:
@atymic I think the GeocoderEarth provider is ready to be upgraded (using the Pelias provider). Could you review it ? :)

jbelien:
There is no breaking changes so I guess we can release this as `1.1.0`, right ?

Ping @Nyholm

Nyholm:
If the Public API did not change than this is fine. 

Why did we need to update all the cached test responses?

jbelien:
> Why did we need to update all the cached test responses?

It was indeed probably not necessary.
Would you prefer me to revert it ?

Nyholm:
Na, thats fine. I just wondered. If we have to rewrite all the tests it may be an indicator that we changed the behaviour.

jbelien:
Just to be sure, I just checked.
Update of the cached response was not necessary indeed.

jbelien:
New release `1.1.0`is live : https://packagist.org/packages/geocoder-php/geocode-earth-provider

PR #1011, opened 2019-09-11
3.x

azraeldk:
What about adding such provider as CENSUS.GOV?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,10 +13,11 @@
     ],
     "require": {
         "php": ">=5.4.0",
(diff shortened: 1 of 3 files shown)


atymic:
I'm not sure what you are trying to do here.
If you'd like to suggest a provider, create an issue rather than a PR.

jbelien:
@atymic is right !
This PR is trying to merge branch `3.x` for no reason, I'll close it.

@azraeldk please open a new issue with this (and https://geocoding.geo.census.gov/ link 😉)

PR #1016, opened 2019-09-23
Fix code example

ratibus:
Key and app-id were inverted in the code example vs constructor signature in the implementation.
A quote was missing in another line.


atymic:
Good catch, thanks for the PR :)

PR #1019, opened 2019-09-27
Adding BC checks on every repository

Nyholm:
Im not sure if this is a good idea or not... 
We will not notice this on PullRequests anyways..


atymic:
Is there a reason why it's done on each individual repo rather than when merging here (on the main one)?

Nyholm:
Yes. Because if one package breaks BC the main build will be red. 

It is fine that one package breaks BC (it might prepare for a new major release) but it is never okey for the main build to be red.

PR #1026, opened 2019-10-17
OpenRouteService provider

gaelg:
Hi, here's a provider for OpenRouteService, heavily based on the GeocodeEarth provider which uses the same backend tool: Pelias.

Openrouteservice is managed by University of Heidelberg. They say "Openrouteservice is free and will stay free - for everyone": https://openrouteservice.org/plans/

It's needed for a contribution to Drupal: https://www.drupal.org/project/geocoder/issues/3085549

Please forgive my mistakes, I'm not very used to GitHub contributions.


src/Provider/OpenRouteService/OpenRouteService.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function geocodeQuery(GeocodeQuery $query): Collection
+    {
+        $url = $this->getGeocodeQueryUrl($query, [
+            'api_key' => $this->apiKey,
+        ]);
+
+        $url = str_replace('/v1/', '/', $url);
```
atymic: I'm curious as to the reason this is here. I'm guessing your implementation of pelias doesn't have versioning?

Maybe better to make a change to the base class to allow a null version instead of this (which is a bit hacky)

README.md
```diff
@@ -143,6 +143,7 @@ Provider       | Package | Features | Stats
...
+[OpenRouteService](https://github.com/geocoder-php/open-route-service-provider) | `geocoder-php/open-route-service-provider` | address, reverse <br> [Website](https://openrouteservice.org/) | [![Latest Stable Version](https://poser.pugx.org/geocoder-php/open-route-service-provider/v/stable)](https://packagist.org/packages/geocoder-php/open-route-service-provider) <br>[![Total Downloads](https://poser.pugx.org/geocoder-php/open-route-service-provider/downloads)](https://packagist.org/packages/geocoder-php/open-route-service-provider)
 [Pelias](https://github.com/geocoder-php/pelias-provider) | `geocoder-php/pelias-provider` | address, reverse <br> [Website](https://pelias.io/) | [![Latest Stable Version](https://poser.pugx.org/geocoder-php/pelias-provider/v/stable)](https://packagist.org/packages/geocoder-php/pelias-provider) <br>[![Total Downloads](https://poser.pugx.org/geocoder-php/pelias-provider/downloads)](https://packagist.org/packages/geocoder-php/pelias-provider)
```
atymic: Maybe we should remove this from the readme as it can't actually be used on it's own?

src/Provider/OpenRouteService/OpenRouteService.php
jbelien: According to the documentation, the current version (`v2`) has `/v2/` in the url (version 1 doesn't)  !
Please switch your code to current version (`v2`) !

README.md
jbelien: @atymic What do you mean by "it can't actually be used on it's own" ?

atymic: Sorry, ignore me. I thought we left the underlying pelias provider as abstract, but we didn't.

src/Provider/OpenRouteService/OpenRouteService.php
atymic: Here's some clarification from the linked drupal PR:

> ORS is quite as Pelias but with some variations : API key required, and no version number in API calls. I found no easy way to handle this variations in a Drupal patch, as I cannot implement a subclass and don't want to duplicate code (I could try to use composer exclude-from-classmap to prevent autoloading, but it's weird and not guaranteed to work well).
Actually, the right way to work around the dependency problem is to provide the provider outside from this Drupal module, by contributing directly to geocoder-php, which I did: https://github.com/geocoder-php/Geocoder/pull/1026

> The good side is that anyone not using Drupal will still be able to use my code to interact with OpenRouteService. The bad side is that this feature can't be properly completed before the new provider is reviewed and included in geocoder-php. Then it will be very easy to implement the Drupal adapter.

So it sounds like it actually doesn't have a version, so maybe worth updating the base class to make it optional?

atymic: Looking at the docs here:
https://openrouteservice.org/dev/#/api-docs/geocode/search/get

It seems to not have a version, while most of the other routes do?

jbelien: Oh sorry, you're right. I looked at v2 documentation but didn't notice the geocoding requests are without a ´/v2´ 🤦‍♂ 

I don't know how to handle this the best way.
Pelias does provide a version number ! So, I don't think it's the best way to allow to not setup a version number in Pelias "provider" !

We first need to know which version of Pelias they use in OpenRouteService (and know if they're planning to change/upgrade it in the future).

gaelg: Hi, I asked ORS about the version number: https://ask.openrouteservice.org/t/pelias-version-in-api-url/1021?u=gaelg
And yes, version number is part of the Pelias standard, that's why I made such a hack in the subclass to remove it.
As v1 in the only version available for now, do we need to wait for their response or could we just get this in, and then update the code if needed when/if we get an answer?

atymic: I'd prefer to update the base pelias class to make the `$version` nullable.

I think it's a lot less hacky than making the change here.

https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/Pelias/Pelias.php#L44-L49

Something like this 
```php
public function __construct(HttpClient $client, string $root, ?int $version = 1)
    {
		$this->root = $version !== null
             ? sprintf('%s/v%d', rtrim($root, '/'), $version);
             : $root


        $this->version = $version;
        parent::__construct($client);
    }

Comment (jbelien) [reply src/Provider/OpenRouteService/OpenRouteService.php]: Haha, I'd prefer the opposite ; not to touch the Pelias provider because the version number is part of Pelias !

@Nyholm What do you think ?

Comment (atymic) [reply src/Provider/OpenRouteService/OpenRouteService.php]: Fair enough haha.

Happy to defer to @Nyholm

Comment (Nyholm) [reply src/Provider/OpenRouteService/OpenRouteService.php]: Oh, interesting. 
@jbelien says that version should be a part of the url according to Pelias specification. But the OpenRouteService has decided not to do that. That is why we have to write this little hack with `str_replace`, correct?

Are there any other Pelias providers that need this hack? Eg: Is OpenRouteService a super special thing or does this happens a lot? If it happens a lot I would consider changing our abstraction (`Pelias` provider). 

If this is our only current case, I would prefer if we kept this hack here, or preferably, add/update the `OpenRouteService` to avoid the `str_replace`. 

Does it make sense?

Comment (jbelien) [reply src/Provider/OpenRouteService/OpenRouteService.php]: > That is why we have to write this little hack with str_replace, correct?

Yes!

> Are there any other Pelias providers that need this hack?

AFAIK, no other Pelias provider do that!

> If this is our only current case, I would prefer if we kept this hack here, or preferably, add/update the OpenRouteService to avoid the str_replace.

I didn't get that part. I would prefer to keep the special case locally to OpenRouteService provider (and not touch the Pelias provider).
But the `str_replace` is proposed in this OpenRouteService PR.
Should we find a "better" hack than `str_replace` ? Is it what you mean ?


Thanks :)

Comment (Nyholm) [reply src/Provider/OpenRouteService/OpenRouteService.php]: Oh, Im missing a word.. =)

> "add/update the OpenRouteService's constructor to avoid the str_replace."

Ie write a new value to `$this->root`

Comment (jbelien) [reply src/Provider/OpenRouteService/OpenRouteService.php]: Yes! That would work indeed! 🎉 👍

Comment (gaelg) [reply src/Provider/OpenRouteService/OpenRouteService.php]: Like this? https://github.com/gaelg/Geocoder/commit/1118bbb8256f15e2375ca58133e99a8c59b4a020

File: src/Provider/OpenRouteService/OpenRouteService.php [hunk 3]
```diff
@@ -0,0 +1,89 @@
...
+        $this->apiKey = $apiKey;
+        parent::__construct($client, self::API_URL, self::API_VERSION);
+
+        /*
+         * Openrouteservice does not use /v1 in first version, but plan to add
+         *  /v2 in next version.
+         *
+         * @see https://ask.openrouteservice.org/t/pelias-version-in-api-url/1021
+         */
+        if (self::API_VERSION == 1) {
```
atymic: strict comparsion please

atymic:
Are we good to go on this PR @jbelien?

src/Provider/OpenRouteService/OpenRouteService.php
```diff
@@ -0,0 +1,89 @@
...
+
+        return $this->executeQuery($url);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName(): string
+    {
+        return 'open_route_service';
```
jbelien: I would put `openrouteservice` (without underscores) since it is how they refer to it (see https://github.com/GIScience/openrouteservice)

src/Provider/OpenRouteService/README.md
```diff
@@ -0,0 +1,27 @@
+# OpenRouteService Geocoder provider
+[![Build Status](https://travis-ci.org/geocoder-php/open-route-service-provider.svg?branch=master)](http://travis-ci.org/geocoder-php/open-route-service-provider)
```
jbelien: I would put openrouteservice (without dashes) since it is how they refer to it (see https://github.com/GIScience/openrouteservice)

src/Provider/OpenRouteService/README.md
```diff
@@ -0,0 +1,27 @@
...
+[![Quality Score](https://img.shields.io/scrutinizer/g/geocoder-php/open-route-service-provider.svg?style=flat-square)](https://scrutinizer-ci.com/g/geocoder-php/open-route-service-provider)
+[![Software License](https://img.shields.io/badge/license-MIT-brightgreen.svg?style=flat-square)](LICENSE)
+
+This is the OpenRouteService provider from the PHP Geocoder. This is a **READ ONLY** repository. See the
+[main repo](https://github.com/geocoder-php/Geocoder) for information and documentation. 
+
+### Install
+
+```bash
+composer require geocoder-php/open-route-service-provider
```
jbelien: I would put openrouteservice (without dashes) since it is how they refer to it (see https://github.com/GIScience/openrouteservice)

src/Provider/OpenRouteService/composer.json
```diff
@@ -0,0 +1,47 @@
+{
+    "name": "geocoder-php/open-route-service-provider",
```
jbelien: I would put openrouteservice (without dashes) since it is how they refer to it (see https://github.com/GIScience/openrouteservice)

src/Provider/Pelias/README.md
```diff
@@ -21,6 +21,7 @@ You can run your own Pelias instance on your server, see [documentation](https:/
 For instance:
 
 - Geocode Earth ([Website](https://geocode.earth/) & [Provider](https://github.com/geocoder-php/geocode-earth-provider))
+- OpenRouteService ([Website](https://openrouteservice.org/) & [Provider](https://github.com/geocoder-php/open-route-service-provider))
```
jbelien: I would put openrouteservice (without dashes) since it is how they refer to it (see https://github.com/GIScience/openrouteservice)

jbelien:
Seems good to be merged 👍 
I'll take care of the merge and the release of the new provider in the next few days ; thanks a lot !

jbelien:
I was planning to release it today but I miss one admin access to finish the process.
I contacted @Nyholm. Stay tuned ! 👍

gaelg:
@Nyholm any news?

gaelg:
https://github.com/geocoder-php/openrouteservice-provider

atymic:
cc @Nyholm

jbelien:
New provider is released: https://packagist.org/packages/geocoder-php/openrouteservice-provider (thanks for the access @Nyholm)

@gaelg The build is [currently failing](https://travis-ci.org/geocoder-php/openrouteservice-provider), could you have a look at it ?

atymic:
@jbelien the issue is you need to tag the new version of the Pelias provider
https://github.com/geocoder-php/pelias-provider

(old version has root as protected, so the queries are failing because it can't set the correct root URL)

jbelien:
> 
> 
> @jbelien the issue is you need to tag the new version of the Pelias provider
> https://github.com/geocoder-php/pelias-provider
> 
> (old version has root as protected, so the queries are failing because it can't set the correct root URL)

Oh, right, I'll take care of that tonight! Thanks!

jbelien:
@atymic You were right, I released the new version of Pelias provider and the OpenRouteService provider build is now okay! Thanks :)

atymic:
No worries :)

PR #1027, opened 2019-10-18
Exception interface should extend Throwable interface

maxhelias:
PhpStan warns of an error about this

```
124    PHPDoc tag @throws with type Geocoder\Exception\Exception is not subtype of Throwable
```

And this has no impact as each exception inherits from \Exception

### DIFF
--- a/src/Common/Exception/Exception.php
+++ b/src/Common/Exception/Exception.php
@@ -15,6 +15,6 @@ namespace Geocoder\Exception;
 /**
  * @author William Durand <william.durand1@gmail.com>
  */
-interface Exception
+interface Exception extends \Throwable
 {
 }


maxhelias:
friendly ping @Nyholm :)

jbelien:
> But we can just extend it as the minimum supported version in php 7.

@atymic What doy you mean by this ?

maxhelias:
I think he means that it's the same thing as Guzzle PR but without managing compatibility because the minimum requirement is PHP 7

jbelien:
Oh okay, that makes sense!
Then I don't see any reason not to merge this :)

maxhelias:
@jbelien Can we move forward and merge it ?

maxhelias:
Thanks @jbelien :)

Issue #1028, opened 2019-10-19
API access denied.


klikpaspor:
` Message: You must use an API key to authenticate each request to Google Maps Platform APIs. For additional information, please refer to http://g.co/dev/maps-no-account`

```
use Geocoder\Query\GeocodeQuery;
use Geocoder\Query\ReverseQuery;

$httpClient = new \Http\Adapter\Guzzle6\Client();
$provider = new \Geocoder\Provider\GoogleMaps\GoogleMaps($httpClient);
$geocoder = new \Geocoder\StatefulGeocoder($provider, 'en');

$result = $geocoder->geocodeQuery(GeocodeQuery::create('Buckingham Palace, London'));
$result = $geocoder->reverseQuery(ReverseQuery::fromCoordinates(...));
```
API access denied.

atymic:
As per the link in the error, you need to use an API key.

https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/GoogleMaps/GoogleMaps.php#L82-L89

We should probably clarify this is the docs, though.

klikpaspor:
version 3.x no need an Api key right? or Google has updated?

atymic:
Sorry about the accidental close. 
PR to update the docs here: https://github.com/geocoder-php/Geocoder/pull/1030

PR #1030, opened 2019-10-25
docs: update google maps code in readme

atymic:
Google maps has required an api key for fairly long time, but it's missing from the documentation (which might confuse people copy/pasting).

I've added the api key to the params in the code example

Closes https://github.com/geocoder-php/Geocoder/issues/1028

### DIFF
--- a/src/Common/Tests/ProviderAggregatorTest.php
+++ b/src/Common/Tests/ProviderAggregatorTest.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 namespace Geocoder\Tests;
 
 use Geocoder\Collection;
-use Geocoder\Geocoder;
 use Geocoder\Model\Address;
 use Geocoder\Model\AddressCollection;
 use Geocoder\Query\GeocodeQuery;
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ use Geocoder\Query\GeocodeQuery;
 use Geocoder\Query\ReverseQuery;
 
 $httpClient = new \Http\Adapter\Guzzle6\Client();
-$provider = new \Geocoder\Provider\GoogleMaps\GoogleMaps($httpClient);
+$provider = new \Geocoder\Provider\GoogleMaps\GoogleMaps($httpClient, null, 'your-api-key');
 $geocoder = new \Geocoder\StatefulGeocoder($provider, 'en');
 
 $result = $geocoder->geocodeQuery(GeocodeQuery::create('Buckingham Palace, London'));
--- a/src/Provider/GoogleMaps/GoogleMaps.php
+++ b/src/Provider/GoogleMaps/GoogleMaps.php
@@ -47,7 +47,7 @@ final class GoogleMaps extends AbstractHttpProvider implements Provider
     private $region;
 
     /**
-     * @var string|null
+     * @var string
      */
     private $apiKey;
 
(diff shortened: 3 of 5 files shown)


jbelien:
That's a good idea but I think we should also update the GoogleMaps provider to make the API key **not** optional!

Caution: in your example you used `$privateKey` but I think the API key goes to `$apiKey`

atymic:
> That's a good idea but I think we should also update the GoogleMaps provider to make the API key not optional!

Good idea. I've updated the constructor to throw an exception if you try to construct without an api key since we can't remove the default of null without breaking BC.

> Caution: in your example you used $privateKey but I think the API key goes to $apiKey

Not sure what you mean by this? The 3rd constructor arg is `apiKey`:
https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/GoogleMaps/GoogleMaps.php#L98-L103

atymic:
Hmm, not sure how the best way to throw an exception is, as the business API uses a private key which isn't added until after the provider is constructed.

jbelien:
> Not sure what you mean by this? The 3rd constructor arg is apiKey:
https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/GoogleMaps/GoogleMaps.php#L98-L103

Shit, sorry, I was looking at the business constructor ...

> Hmm, not sure how the best way to throw an exception is, as the business API uses a private key which isn't added until after the provider is constructed.

Looking at the parameters, the business constructor uses both `privateKey` and `apiKey` (5th argument).
So I think we could throw an exception in `__construct()` if there is no API key.

atymic:
> So I think we could throw an exception in __construct() if there is no API key.

The issue with that is when using business, you can do so without an api key (private key only) :(

jbelien:
Damn, didn't know about that!

Do you think it would make sense to separate Google Maps and Google Maps Business in 2 separated providers ?

atymic:
> Damn, didn't know about that!
> Do you think it would make sense to separate Google Maps and Google Maps Business in 2 separated providers ?

The issue is that also causes a BC break. 

I think the best way is probably to add a 4th constructor param, `isBusiness` to skip the API key check. It's a little bit hacky but that way we can preserve BC.

atymic:
cc @Nyholm for thoughts :)

Nyholm:
Im. I see where you are coming from. 
The API made some BC breaking changes and now we want to reflect those changes in the client without breaking BC... 

The way I see it there are two clean (not hacky) solutions. 
1) Just update documentation and comments. No code changes
2) Accept that we need to do a new major release and maybe also do more work on this client. 

I would prefer (1) because a new major version does not give current users a good reason to upgrade. Ie no new features it is just a code clean up.

But that is just my 2 cents.

atymic:
Thanks for your thoughts @Nyholm 

I've updated the documentation for the provider to make it a lot clearer, and I've  made it so it will throw an exception if you try to geocode without either an API key or a business client it.

src/Common/Tests/ProviderAggregatorTest.php
```diff
@@ -13,7 +13,6 @@
 namespace Geocoder\Tests;
 
 use Geocoder\Collection;
-use Geocoder\Geocoder;
```
atymic: StyleCI made me do this.

PR #1034, opened 2019-12-18
GoogleMapsPlaces: check if value "open_now" in "opening_hours"

bitecodes:
If you perform the following request you will see that the value of "opening_hours" is an empty object and "open_now" is not set.

https://maps.googleapis.com/maps/api/place/textsearch/json?location=51.0572773,13.7763207&rankby=distance&type=transit_station&key=YOUR_API_KEY

```javascript
{
  "results": [
    {
      "id": "b427385cd417b5851b4e23de4225d102ffc2aa43",
      "opening_hours": {},
      // ...
    },
    // ...
  ],
  // ...
}
```
This will result in the following error.

```
In OpeningHours.php line 76:

  Notice: Undefined property: stdClass::$open_now
```
Therefore, we should check if the property "open_now" exists.

### DIFF
--- a/src/Provider/GoogleMapsPlaces/Model/OpeningHours.php
+++ b/src/Provider/GoogleMapsPlaces/Model/OpeningHours.php
@@ -20,7 +20,7 @@ use stdClass;
 class OpeningHours
 {
     /**
-     * @var bool
+     * @var bool|null
      */
     private $openNow;
 
@@ -35,11 +35,11 @@ class OpeningHours
     private $weekdayText;
 
     /**
-     * @param bool     $openNow
-     * @param array[]  $periods
-     * @param string[] $weekdayText
+     * @param bool|null $openNow
+     * @param array[]   $periods
+     * @param string[]  $weekdayText
      */
-    public function __construct(bool $openNow, array $periods, array $weekdayText)
+    public function __construct($openNow, array $periods, array $weekdayText)
     {
         $this->openNow = $openNow;
         $this->periods = $periods;
@@ -47,9 +47,9 @@ class OpeningHours
     }
 
     /**
-     * @return bool
+     * @return bool|null
      */
-    public function isOpenNow(): bool
+    public function isOpenNow()
     {
         return $this->openNow;
     }
@@ -70,10 +70,10 @@ class OpeningHours
         return $this->weekdayText;
     }
 
-    public static function fromResult(Stdclass $openingHours): self
+    public static function fromResult(stdClass $openingHours): self
     {
         return new self(
-            $openingHours->open_now,
+            $openingHours->open_now ?? null,
             $openingHours->periods ?? [],
             $openingHours->weekday_text ?? []
         );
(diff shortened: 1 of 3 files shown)


atymic:
Thanks for the PR @bitecodes 

Can you please add a test to cover the example you provided?

src/Provider/GoogleMapsPlaces/GoogleMapsPlaces.php
```diff
@@ -348,7 +348,7 @@ private function fetchUrl(string $url, array $query): AddressCollection
                 $address = $address->withWebsite($result->website);
             }
 
-            if (isset($result->opening_hours)) {
+            if (isset($result->opening_hours->open_now)) {
```
atymic: Could we make this change in the model please (ie default to false if no hours)?
https://github.com/geocoder-php/Geocoder/blob/master/src/Provider/GoogleMapsPlaces/Model/OpeningHours.php#L76

Thanks!

bitecodes: But wouldn't this  result in a false assumption? Then the user would assume that the place is closed, but in reality we can not make an assertion. The place could be closed or open. Also setting `open_now` to `null` would probably throw off most users, because they would not check if the value is a boolean.

For me personally an empty object or null should be treated the same, as both values imply that no information about the opening hours exist.

bitecodes:
> Thanks for the PR @bitecodes
> 
> Can you please add a test to cover the example you provided?

Yes, I will add a test

src/Provider/GoogleMapsPlaces/GoogleMapsPlaces.php
atymic: Fair point. 
Is there any cases where `open_now` isn't set but the opening hours are?

bitecodes: Hmm, it seems to be a bit more complicated.

According to the docs, the Text Search Request and the Nearby Search Request only return the field `open_now`. There are no fields `periods` and `weekday_text` for these requests. I could also not find any results that did include any fields other than `open_now`.

https://developers.google.com/places/web-service/search#PlaceSearchResults

The Find Place Request on the other hand might also return a field `weekday_text`. It is only show in the JSON Response, but they did not specify what is returned in the docs. I could not find any information about a field of `periods`, so I don't know where that comes from.

https://developers.google.com/places/web-service/search#find-place-responses

So I don't know if there are cases where `open_now` is not set, but `weekday_text` exists.

Should we add a `hasOpenNowInformation` method to `OpeningHours` and allow `isOpenNow` to be nullable? What would you suggest?

atymic: I would just make it nullable, and add a note to the docblock mentioning that it will return null if there's no information. As long as we add the correct docblocks, it should show the return types etc in IDEs.

bitecodes:
Added a test and made open_now nullable

bitecodes:
Code style fixed! ;-)

atymic:
Merged 🎉 

Could you please tag a new version of the package @Nyholm?
I only have permissions on this repo, not on the subtree split versions.

Thanks!

jbelien:
Done: [`1.0.1`](https://github.com/geocoder-php/google-maps-places-provider/releases/tag/1.0.1) 🎉

atymic:
Thanks @jbelien :)

PR #1038, opened 2020-01-08
[LocationIQ] Normalize City

### LINKED ISSUE #1037: Cannot get city with LocationIQ provider
Hi,

I use the [LocationIQ provider](https://github.com/geocoder-php/locationiq-provider) and I'm unable to retrieve the city when I get the `Location` object by calling `geocodeQuery` method for the following address: '12 rue Beziel' (french address). 

So I checked the raw response for this address in the LocationIQ API  geocode endpoint url (doc: https://locationiq.com/docs-html/index.html#search-forward-geocoding) and I got this for the address details:
```
{
  "house_number": "12",
  "road": "Rue Béziel",
  "town": "Rambouillet",
  "county": "Rambouillet",
  "state": "Île-de-France",
  "country": "France",
  "postcode": "78120",
  "country_code": "fr"
}
```
It seems some properties are missing with this address.

In [LocationIQ.php](https://github.com/geocoder-php/locationiq-provider/blob/master/LocationIQ.php) file, `town` and `county` for example are not used when the Location object is built.
There is a way to have a fallback if the `city` property name is missing ?

Many thanks!

jbelien:
Add `normalizecity` parameter in the request to make sure there is a `city` property in the response.
See https://locationiq.com/docs-html/index.html#locationiq-specific-parameters

Close #1037


jbelien:
@pixrr Could you confirm this fixes your "issue" ?

pixrr:
> @pixrr Could you confirm this fixes your "issue" ?

Yes, it seems good :+1:

jbelien:
> Edit 2: It does not seem to work with the xml format 🤔

Good catch!
They mention it somehow in the [documentation](https://locationiq.com/docs-html/index.html#search-forward-geocoding):
> This version (v1) of our Search API is compatible with OpenStreetMap's Nominatim Geocoder in both JSON & XML formats. However, all our enhancements such as additional datasets and algorithms are supported only in `json` or `xmlv1.1` format options.

Issue #1035, opened 2020-01-07
How to limit API calls per second


tmp-hallenser:
Hi,

I want to use the Nominatim API to do some geocoding. According to the Usage Policy, one is allowed to do max. 1 request per second. 

Since several calls will be identical, I'm using the Cache provider to reduce the number of calls. 

Is there a smart way to achieve the 1 request per second restriction in case of caching?

I could of course just pause for 1s after each query. This is a bit suboptimal since it unnecessarily slows down cached queries.

Many thanks!

atymic:
I think the easiest way to do this would be to using some kind of middleware in the HTTP layer.
You could write a HTTplug plugin, but if you're using guzzle there's a preexisting plugin: 
https://github.com/hamburgscleanest/guzzle-advanced-throttle

Maybe give that a go?

tmp-hallenser:
Thanks. Here's a working example using https://github.com/spatie/guzzle-rate-limiter-middleware

It combines caching with rate limiting, i.e. limits calls to API but first checks the cache.

Perhaps it's worth adding a comment to the readme.

```
<?php
// https://github.com/spatie/guzzle-rate-limiter-middleware
require __DIR__ . '/vendor/autoload.php';

$stack = \GuzzleHttp\HandlerStack::create();
$stack->push(\Spatie\GuzzleRateLimiterMiddleware\RateLimiterMiddleware::perMinute(1));

$httpClient = new \GuzzleHttp\Client(['handler' => $stack, 'timeout' => 30.0, ]);
$httpAdapter = new \Http\Adapter\Guzzle6\Client($httpClient);
$psr6Cache = new \Cache\Adapter\PHPArray\ArrayCachePool();
$provider = new \Geocoder\Provider\Nominatim\Nominatim($httpAdapter, 'https://nominatim.openstreetmap.org', 'Geocoder test');
$cachedProvider = new \Geocoder\Provider\Cache\ProviderCache($provider, $psr6Cache);
$geocoder = new \Geocoder\StatefulGeocoder($cachedProvider, 'en');
$formatter = new \Geocoder\Formatter\StringFormatter();

$gps_coords = array(
    array(
        'lat' => 52.516275,
        'lgn' => 13.377704
    ) ,
    array(
        'lat' => 51.503396,
        'lgn' => - 0.12764
    ) ,
    array(
        'lat' => 52.516275,
        'lgn' => 13.377704
    )
);
foreach ($gps_coords as $gps_coord)
{
[...]

PR #1039, opened 2020-01-12
feat: add rate limiting cookbook documentation

atymic:
This PR adds the example from #1035 into the cookbook as documentation as it seems like useful information :)


jbelien:
Excellent idea to add this to our documentation!
Thanks @atymic ! 🎉

PR #1042, opened 2020-01-25
feat: support api key on here provider

### LINKED ISSUE #1041: Here Provider now uses an API Key for REST
Would it be possible to get support for the API KEY authentication in the Here provider? It uses a different URL and only needs the API KEY attribute.

atymic:
I've updated the provider so it can use both the API key and appcode/appkey.
The issue is that's a bit of a mess. Since the appcode/appkey is deprecated, maybe we should consider bumping the major version and dropping support for the old style auth?

Closes #1041


jbelien:
Thanks @atymic ! 👍 
I'll review it but could you make sure it passes the tests first.

AppCode doesn't seem to be completely deprecated, they are still usable but new users apparently can't create it anymore (see https://developer.here.com/documentation/authentication/dev_guide/topics/app-credentials.html)
Therefore I would prefer to keep both options possible.
What's your opinion @Nyholm ?

atymic:
> I'll review it but could you make sure it passes the tests first.

They actually do all pass locally, something is wonky with CI.
Will try to figure it out

> AppCode doesn't seem to be completely deprecated, they are still usable but new users apparently can't create it anymore (see developer.here.com/documentation/authentication/dev_guide/topics/app-credentials.html)
Therefore I would prefer to keep both options possible.

While I did implement it with both options, on second thoughts I don't think there's much point it having both. Anyone still on the old auth method can switch over via the dashboard pretty easily, and nothing is going to break for existing users (due to the major version bump)

Anyone new coming to the package can use the next major version with the API key.

atymic:
Ok, i've tried redoing the cache files with no luck. Any ideas why they are not working in CI?

image

alteredorange:
So is Here not currently working? I just signed up, and can't get app code credentials, but when I just put in an API key, it says I need an app code...

pfuhrmann:
Would be great to get this merged. The current state makes `here-provider` unusable for any new user or any user that needs to refresh the credentials.

@jbelien do you have ideas about why CI is failing? Can I help somehow to get this merged @atymic ?

jbelien:
> @jbelien do you have ideas about why CI is failing? Can I help somehow to get this merged @atymic ?

I'm having the same issue locally ... Investigating it now :)

src/Provider/Here/Readme.md
```diff
@@ -8,23 +8,26 @@
...
 
 ```bash
 composer require geocoder-php/here-provider
 ```
 
-### App Id and App Code
+## Using
 
-Get your Here credentials at https://developer.here.com/
+TODO
```
jbelien, line 24: Add documentation on how to use it (API key vs AppId/AppCode) !

jbelien:
@atymic Could you add some documentation and merge the Style CI fix we did in `master` ? Thanks.

atymic:
On it :)

atymic:
@jbelien docs added and style issues resolved :)

jbelien:
Awesome! Thanks @atymic 🎉

jbelien:
Version `0.5.0` is [released](https://packagist.org/packages/geocoder-php/here-provider).

PR #1046, opened 2020-02-10
[Pelias] Fix NULL value for $bounds

### LINKED ISSUE #1043: [Pelias] Notice: Trying to access array offset on value of type null on php 7.4
Hi
Got error on **php 7.4** 
**Notice: Trying to access array offset on value of type null on php**

in vendor/willdurand/geocoder/Model/Address.php (line 258)

jbelien:
In `Address::createFromArray()` function, `$bound` needs to be an array !

Fix #1043


src/Provider/Pelias/Pelias.php
```diff
@@ -166,7 +166,12 @@ protected function executeQuery(string $url): AddressCollection
                     'east' => $location['bbox'][0],
                 ];
             } else {
-                $bounds = null;
+                $bounds = [
```
atymic: maybe worth adding a test case to cover this bug?

jbelien: Good idea! 👍

jbelien: Test added but I don't have an API key to add the cached response (I requested one).

jbelien:
@atymic Would you have time to review my last changes, so I can merge this ? Thanks

PS: I'll fix CI (failed) test in `master` (since it's not related to this PR) and Travis builds passed (not sure why it still shows as pending here)

PR #1051, opened 2020-02-28
[Nominatim] Fix Undefined property when no country

AntoineLemaire:
I got a case where OpenStreetMap does not return Country nor CountryCode :)


src/Provider/Nominatim/Tests/NominatimTest.php
```diff
@@ -180,4 +180,23 @@ public function testGeocodeNoOSMId()
...
+
+        $result = $results->get(1);
+        $this->assertInstanceOf('\Geocoder\Model\Address', $result);
+        $this->assertEquals('Data © OpenStreetMap contributors, ODbL 1.0. https://osm.org/copyright', $result->getAttribution());
+
+        $this->assertEquals('Italia', $result->getDisplayName());
+        $this->assertEquals('waterway', $result->getCategory());
+        $this->assertEquals('62194430', $result->getOSMId());
+        $this->assertEquals('way', $result->getOSMType());
+        $this->assertEquals(null, $result->getCountry());
```
jbelien, line 200: Could you also add an asset for the country code ?
Thanks

jbelien:
@AntoineLemaire Could you add the assert for the country code so we can merge this ? Thanks !

src/Provider/Nominatim/Tests/NominatimTest.php
jbelien: Oh damn ! You're absolutely right !
Sorry :P

PR #1053, opened 2020-04-05
add storage location provider with tests and documentation

apacheborys:
Tried to create provider what can make job without calling to any 3-rd party service.
Benefits:
* save requests to real provider
* own-driven performance control
* opportunity to build own locations, places
* possible to use high precise value for coordinates (storing as float type)


jbelien:
Hello @apacheborys, thanks for your contribution.

I'm a bit confused though.
I have the feeling you try to add 2 features in 1:

1. Add the ability to generate (and use) your own database of addresses ;
2. Add polygon abilities ;

The first one seems interesting.
I have some "concerns" about the second one. So far, the whole geocoder library is based on [Address](https://github.com/geocoder-php/Geocoder/blob/master/src/Common/Model/Address.php) as [points](https://github.com/geocoder-php/Geocoder/blob/master/src/Common/Model/Coordinates.php) and it seems strange to extend it to polygon.

That's my opinion though ; what do you thing @Nyholm @atymic ?

apacheborys:
Hello @jbelien ,

Thank you for your feedback. Sure I understand your confusion. In this provider I tried to reproduce my experience in saving requests to real provider.

Yes, it's little bit another direction of concept, but probably you can find it useful for Geocoder? Also, in the future I think will be better - to extend Address entity for locale. This extended entity will make possibility to store locally whole Place entity as well - for all required languages.

Probably you have another vision? I am open to propositions and collaboration.

apacheborys:
@jbelien, in last release I escaped from Address dependency in Place entity. 
Currently it's Address collection inside Place. Additionally I added functionality what able to use with different locales. So, what conclusion will be from your side? Is it valuable Provider for Geocoder?
If talk about provider's benefits, additionally you can use that provider with high precise coordinates. Precision limits only by float type possibilities.

PR #1065, opened 2020-06-25
[Google Maps] Added missing nullable return type

jackff87:
Sometimes Google Maps does not return a Plus Code.
PoC:
{"query":"[object] (Geocoder\\Query\\GeocodeQuery: GeocodeQuery: {\"text\":\"Port of Ijmuiden, North Holland, Netherlands, Port Ijmuiden Port\",\"bounds\":\"null\",\"locale\":null,\"limit\":5,\"data\":{\"mode\":\"search\",\"fields\":\"formatted_address, geometry, icon, name, place_id, plus_code, types\"}})"} 

The first 2 results have a PlusCode but the rest do not, which generates an exception.

### DIFF
--- a/src/Provider/GoogleMapsPlaces/Model/GooglePlace.php
+++ b/src/Provider/GoogleMapsPlaces/Model/GooglePlace.php
@@ -194,7 +194,7 @@ final class GooglePlace extends Address
     /**
      * @return PlusCode|null
      */
-    public function getPlusCode(): PlusCode
+    public function getPlusCode(): ?PlusCode
     {
         return $this->plusCode;
     }


jbelien:
Hello @jackff87 ,

Thanks for your PR !
You're right but currently our library supports PHP 7.0 and nullable types only appear in PHP 7.1.

src/Provider/GoogleMapsPlaces/Model/GooglePlace.php
```diff
@@ -194,7 +194,7 @@ public function withIcon(string $icon = null)
     /**
      * @return PlusCode|null
      */
-    public function getPlusCode(): PlusCode
+    public function getPlusCode(): ?PlusCode
```
Nyholm, line 197: This is a bugfix. The docs clearly states what the return type should be.

jbelien:
I'll merge this PR after dropping PHP 7.0 support (see #1066) 👍

jackff87:
@jbelien, @Nyholm  
glad to help :D

PR #1067, opened 2020-06-25
[Mapbox] Check for "short_code" key before using it

damienalexandre:
When using Mapbox to look for places, we may not have a short_code in the contextual part of the response.

For example for Monaco we get this from the API:

```
array:6 [▼
  "id" => "country.3435803554126140"
  "wikidata" => "Q235"
  "text_it" => "Principato di Monaco"
  "language_it" => "it"
  "text" => "Principato di Monaco"
  "language" => "it"
]
```

This PR fix an exception when trying to hydrate such Mapbox results.

> Notice: Undefined index: short_code


src/Provider/Mapbox/Mapbox.php
```diff
@@ -347,7 +347,9 @@ private function updateAddressComponent(AddressBuilder $builder, string $type, a
 
             case 'country':
                 $builder->setCountry($value['text']);
-                $builder->setCountryCode(strtoupper($value['short_code']));
+                if (!empty($value['short_code'])) {
```
jbelien, line 350: Shouldn't you use `isset()` instead of `!empty()` ?

damienalexandre: I guess it depends on the code style Geocoder apply because from a result point of view it's the same. 

I'm switching to `isset` + fixing the tests :+1:

jbelien:
Thanks for your contribution ! 👌

lyrixx:
Hello. Could some tag a new release with this patch? thanks

jbelien:
> Hello. Could some tag a new release with this patch? thanks

Yes, of course !
We're working on another PR (#1068). I'll publish a new release when that PR is merged.
I'll most probably take care of it this weekend !

jbelien:
Version 1.1.0 has been released : https://packagist.org/packages/geocoder-php/mapbox-provider 🎉

lyrixx:
It's very sad, but our application is blocked on PHP 7.0.
I know, it's a shame :( 
Could you tag a 1.0.* version with this patch and without the drop of PHP < 7.2?
:pray: Thanks

jbelien:
I don't see an easy and clean way to make a release at the stage you want.
**but** you should be able to "require" the package at the commit you want.
```
composer require geocoder-php/mapbox-provider@
dev-master#a05cfc3718ccc7a724ddcc392c88727ed62a0509
```

lyrixx:
Okay no problem' thanks

PR #1068, opened 2020-06-26
Drop support for PHP 7.0 and PHP 7.1 (End of life)

### LINKED ISSUE #1066: Drop support for PHP 7.0 (End of life)
PHP 7.0 end of life was more than a year ago (1st January 2019).

Could we drop the support for this version ?
Should we also drop the support for PHP 7.1 (End of life begin of this year) ?

That would allow us to use nullable types, among other things (see #1065).

jbelien:
Close #1066


jbelien:
Build fails on PHP 7.4 !
I'll check (and fix) it this weekend before merging this !

jbelien:
Well, I'm a bit puzzled about why it fails on PHP 7.4.

I guess it's related to
```
Function ReflectionType::__toString() is deprecated
```
but I can replicate it using the same PHP version : `7.4.7`

Any idea @Nyholm @atymic ?

jbelien:
Okay, I think I found it !
It was caused by PHPUnit 6 ; I upgraded it to PHPUnit 7 !

jbelien:
> I like that we normalized the composer.json but it made things a bit hard to review.

Yes, sorry about that ! I should have made 2 separate PR ; thought about it "too late" :P

jbelien:
Here is a summary of the changes (outside formatting/normalization): 

- Remove PHP 7.0
- Remove PHP 7.1
- Add PHP 7.4
- Set minimum PHP version to 7.2 (instead of 7.0)
- Remove `phpunit/phpunit` 6.5
- Upgrade  nyholm/psr7`
- Fix one TomTom provider test

PR #1073, opened 2020-07-04
Improve functions signature

jbelien:
Now that we removed support for PHP 7.0, we can use [nullable types](https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types).

I took the opportunity to improve functions signature.

### DIFF
--- a/src/Common/Model/Country.php
+++ b/src/Common/Model/Country.php
@@ -35,7 +35,7 @@ final class Country
      * @param string $name
      * @param string $code
      */
-    public function __construct(string $name = null, string $code = null)
+    public function __construct(?string $name = null, ?string $code = null)
     {
         if (null === $name && null === $code) {
             throw new InvalidArgument('A country must have either a name or a code');
@@ -50,7 +50,7 @@ final class Country
      *
      * @return string|null
      */
-    public function getName()
+    public function getName(): ?string
     {
         return $this->name;
     }
@@ -60,7 +60,7 @@ final class Country
      *
      * @return string|null
      */
-    public function getCode()
+    public function getCode(): ?string
     {
         return $this->code;
     }
--- a/src/Common/Query/ReverseQuery.php
+++ b/src/Common/Query/ReverseQuery.php
@@ -53,7 +53,7 @@ final class ReverseQuery implements Query
      *
      * @return ReverseQuery
      */
-    public static function create(Coordinates $coordinates)
+    public static function create(Coordinates $coordinates): self
     {
         return new self($coordinates);
     }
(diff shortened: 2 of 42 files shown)


jbelien:
I would like to add `float` type to `createCoordinates()` parameters 
https://github.com/geocoder-php/Geocoder/blob/fce4473c2c7db8d455eb31176282c3eb5ae6ea30/src/Common/Model/Address.php#L279
but I noticed than type is not currently "forced" because some provider send `string` to that function and it's converted in `Coordinates` class:
https://github.com/geocoder-php/Geocoder/blob/fce4473c2c7db8d455eb31176282c3eb5ae6ea30/src/Common/Model/Coordinates.php#L41-L42

What do you think about forcing the type and "fix" the providers than send `string` ?

Nyholm:
Some of these are BC breaks. We dont plan to make a new major release, right?

We should only modify final classes.

jbelien:
> We dont plan to make a new major release, right?

That's the question indeed :P

Nyholm:
I dont think so.. there is no reason to break BC just because of dropping PHP 5. I mean, it should not be the **only** reason.

jbelien:
Okay, I'll let this PR pending then ! 👌

I release a minor release for each provider following the last updated !

src/Common/Model/Country.php
```diff
@@ -35,7 +35,7 @@ final class Country
      * @param string $name
      * @param string $code
```
atymic, line 36: worth updating or removing docblock?

src/Common/Query/ReverseQuery.php
```diff
@@ -64,7 +64,7 @@ public static function create(Coordinates $coordinates)
      *
      * @return ReverseQuery
      */
-    public static function fromCoordinates($latitude, $longitude): self
+    public static function fromCoordinates(float $latitude, float $longitude): self
     {
```
atymic, line 68: Could this break existing code where an int was passed?

src/Provider/BingMaps/BingMaps.php
```diff
@@ -98,7 +98,7 @@ public function getName(): string
      *
      * @return \Geocoder\Collection
      */
-    private function executeQuery(string $url, string $locale = null, int $limit): Collection
+    private function executeQuery(string $url, ?string $locale = null, int $limit): Collection
```
atymic, line 101: can we remove the `null` default since it's got a required param after?

src/Common/Query/ReverseQuery.php
jbelien: See https://github.com/geocoder-php/Geocoder/pull/1073#issuecomment-653751763

atymic: Oops sorry didn't see that comment 🙈

PR #1081, opened 2020-09-23
[Mapbox] Add support for the fuzzyMatch query parameter

DieterHolvoet:
> Specify whether the Geocoding API should attempt approximate, as well as exact, matching when performing searches (`true`, default), or whether it should opt out of this behavior and only attempt exact matching (`false`). For example, the default setting might return `Washington, DC` for a query of `wahsington`, even though the query was misspelled.

https://docs.mapbox.com/api/search/#forward-geocoding


jbelien:
Hello @DieterHolvoet,

We fixed these last few days the Style CI. Could you merge our `master` branch so the Style CI check passes ?
Could you also add a test with the `fuzzyMatch` query parameter ?

Thanks.

DieterHolvoet:
Done! Would you mind adding the hacktoberfest-accepted label to this pull request?

jbelien:
Don't forget to add the cached response.
Could you also add a test with `fuzzyMatch` to `true` ?
Thanks.

DieterHolvoet:
Done!

PR #1091, opened 2020-11-18
Update Komoot TLD.

alexsegura:
As seen on https://photon.komoot.io/

> Note:
Until October 2020 the API was available under photon.komoot.de. Requests still work as they redirected to photon.komoot.io but please update your apps accordingly.


jbelien:
Thanks for the notice @alexsegura 👍

PR #1095, opened 2020-11-29
Update IpInfoDb.php

rtiagom:
Patch to prevent "Uncaught TypeError: Argument 11 passed to Geocoder\Model\Address".

### DIFF
--- a/src/Provider/IpInfoDb/IpInfoDb.php
+++ b/src/Provider/IpInfoDb/IpInfoDb.php
@@ -137,7 +137,7 @@ final class IpInfoDb extends AbstractHttpProvider implements Provider
 
         $timezone = null;
         if (isset($data['timeZone'])) {
-            $timezone = timezone_name_from_abbr('', (int) substr($data['timeZone'], 0, strpos($data['timeZone'], ':')) * 3600, 0);
+            $timezone = timezone_name_from_abbr('', (int) substr($data['timeZone'], 0, strpos($data['timeZone'], ':')) * 3600, 0) ?: null;
         }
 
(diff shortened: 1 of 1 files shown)


jbelien (changes requested):
Not sure this is the best way to tackle this issue.
I would prefer to process correctly the timezone in every cases!

Could you also add a test with an IP in Australia ?

Thanks!

rtiagom:
I agree that this is not the best way to handle this issue but I don't have the time to work on a proper fix right now. At least the patch makes everything work with Geocoder\Model\Address.

rtiagom:
Regarding the test IP, remember that the reason an Australian IP fails right now is because of daylight saving time and that changes in a few months which would make the test inconclusive.

I would recommend applying the current patch right now and open a ticket to find an alternative to timezone_name_from_abbr later on.

jbelien:
> At least the patch makes everything work with Geocoder\Model\Address.

> I would recommend applying the current patch right now and open a ticket to find an alternative to timezone_name_from_abbr later on.

I don't really consider "hiding" the error to be a patch.

> Regarding the test IP, remember that the reason an Australian IP fails right now is because of daylight saving time and that changes in a few months which would make the test inconclusive.

We store/cache the API response, the test will be the same at anytime.

rtiagom:
> I don't really consider "hiding" the error to be a patch.

Ok, fair enough. I don't have the time to work on this.  In our current project we have no need to use the Address timezone field so it doesn't matter what it returns as long it doesn't break which this patch handles it. Feel free to reject the PR.

PR #1101, opened 2020-12-15
Add GitHub Actions

### LINKED ISSUE #1100: Migration to GitHub Actions
Travis CI is closing https://travis-ci.org/ for https://travis-ci.com/ and with that migration it seems like the free tier will be limited to 10000 credits.

I don't exactly understand the pricing model but I think it's time to migrate to GitHub Actions.

**EDIT:** I'm not even sure our Travis CI tests are still running, I triggered https://travis-ci.org/github/geocoder-php/provider-integration-tests/builds/749849954 1 hour ago and it still didn't start.

@Nyholm @atymic  Any concern about migrating from Travis CI to GitHub Actions ?

jbelien:
Close #1100


jbelien:
I've kept it simple for now.

I'll try to add `deps="low"` and `deps="high"` (from current Travis CI setup).

jbelien:
PHP 7.4 with lowest dependencies is failing : https://github.com/geocoder-php/Geocoder/runs/1559772951

I don't really understand why at the moment.

jbelien:
Still to "fix":

- [ ] Find out why `IP2LocationBinary` is failing
- [ ] Wait for `geoip` extension for PHP 8.0 is available

jbelien:
Once this is merged, I'll also add a separate GitHub Actions workflow in each provider for testing in "sub-repositories".

.github/workflows/provider.yml
```diff
@@ -0,0 +1,95 @@
...
+
+jobs:
+  test:
+    name: PHP ${{ matrix.php-version }} / ${{ matrix.provider }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version: ['7.3', '7.4', '8.0']
+        provider:
```
Nyholm, line 17: We could do something like this if we don't want to list all providers:

https://github.com/brefphp/extra-php-extensions/blob/master/.github/workflows/ci.yml#L19

.github/workflows/provider.yml
```diff
@@ -0,0 +1,95 @@
...
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{ matrix.php-version }}
+        extensions: curl
+    - name: Validate composer.json and composer.lock
+      working-directory: ./src/Provider/${{ matrix.provider }}
+      run: composer validate
+    - name: Install dependencies
+      working-directory: ./src/Provider/${{ matrix.provider }}
+      run: composer install --no-progress
```
Nyholm, line 66: We don't have composer.lock in git, right? If so, we can use composer update.

Nyholm (approved):
Well done!

.github/workflows/provider.yml
jbelien: I must admit I'm not sure what the `echo "::set-output name=list::$(ls layers | jq -R -s -c 'split("\n")[:-1]')"` command does 😁

.github/workflows/provider.yml
jbelien: Indeed, we don't commit the `composer.lock` file.
I'll switch to `composer update`. 👍

jbelien:
> Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on. 
> --- Source: https://travis-ci.org/

PR #1102, opened 2020-12-15
Enable PHP 8.0 support

jbelien:
- Add PHP 8.0 support
- Remove PHP 7.2 support ([end of life](https://www.php.net/supported-versions.php))
- Upgrade PHPUnit

I've kept GitHub Actions quite simple.
Migration to GitHub Actions will be handled in #1101


jbelien:
Travis CI is failing with the following error code:
```
PHP Fatal error:  strict_types declaration must be the very first statement in the script in /home/travis/build/geocoder-php/Geocoder/vendor/phpunit/phpunit/phpunit on line 2
```

Since we're moving to GiHub Actions, it's not that big of a deal but I did some research and it's due to the fact that we `require()` PHPUnit script which we are not supposed to do (see https://github.com/sebastianbergmann/phpunit/issues/4096#issuecomment-585900398).

Nyholm (approved):
Awesome. Well done!

PR #1103, opened 2020-12-19
Enable PHP 8.0 support (1)

jbelien:
Following #1102 !

Upgrade PHPUnit for components `Common`, `Http`, and `Plugin`

**Next step:** Release new version of [`willdurand/geocode`](https://github.com/geocoder-php/php-common) and [`geocoder-php/common-http`](https://github.com/geocoder-php/php-common-http) so we can upgrade all providers.

### DIFF
--- a/src/Common/.travis.yml
+++ b/src/Common/.travis.yml
@@ -1,6 +1,11 @@
 language: php
 
-php: 7.2
+matrix:
+    fast_finish: true
+    include:
+        - php: 7.3
+        - php: 7.4
+        - php: 8.0
 
 install:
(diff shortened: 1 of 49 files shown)


Nyholm (approved):
What would the next versions be?

jbelien:
> What would the next versions be?

A new minor version, same logic as when we dropped support for PHP 7.1.

I've included the CHANGELOG files update in this PR.
- `willdurand/geocode` : version `4.4.0`
- `geocoder-php/common-http` : version `4.4.0`
- `geocoder-php/plugin` : version `1.3.0`

Nyholm:
Awesome!

PR #1104, opened 2020-12-21
Enable PHP 8.0 support (2)

jbelien:
Following #1103 !

- Upgrade `php-http/curl-client` for every provider
- Upgrade PHPUnit config

This is the last step to complete full PHP 8.0 support.


Nyholm (approved):
Cool cool 

Great work!

PR #1105, opened 2020-12-28
[ArcGISOnline] Updated Readme to clarify token creation/use.

wantell:
To better clarify the update to the ArcGISOnline Provider I have updated the Readme as follows:
- Explain why there are 2 methods for geocoding
- Provide information about tokens (why/when needed, lifetime, link to instructions for making one)
- Fix some typos

This was prompted by my proposed [update to the Drupal Geocoder module](https://www.drupal.org/project/geocoder/issues/3179963) in support of my previous update to the ArcGISOnline Provider. Since the token has a max lifetime of 14 days, the Drupal code stores the credentials for obtaining a token, and automatically gets a new token when the existing one has expired. However, this generated some confusion due to the `token` method requiring a `token` value while the Drupal UI is requesting values for `client_id` and `client_secret` (the credentials).

Please let me know if you need me to make any further changes to this documentation.

### DIFF
--- a/src/Provider/ArcGISOnline/Readme.md
+++ b/src/Provider/ArcGISOnline/Readme.md
@@ -11,32 +11,46 @@
 This is the ArcGIS provider from the PHP Geocoder. This is a **READ ONLY** repository. See the
 [main repo](https://github.com/geocoder-php/Geocoder) for information and documentation.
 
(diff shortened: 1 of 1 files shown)


jbelien:
Thanks @wantell 🎉

PR #1106, opened 2021-01-14
Add name for PhotonAddress

nihilsen:
This PR adds "name" for the photon provider.

Two new methods are added to `PhotonAddress`: `getName()` and `withName($name)`


jbelien:
Thanks @nihilsen ! 👍 

Could you add a new test for that field ?

nihilsen:
Tests added.

It occurred to me that the Photon API doesn't always provide a "name" property that we can use as the display name.

That's not very useful, so in those cases I think it would be appropriate to construct a placeholder display name using the address components. This produces a display name similar to that given by the Nominatim provider.

jbelien:
> It occurred to me that the Photon API doesn't always provide a "name" property that we can use as the display name.

No, we do not "touch" the result returned by an API.
If there is no `name` property fill, the `name` should stay null.
Please revert commit 4b91de7 and add a test for a request that returns a `name` property.

Thanks.

jbelien (changes requested):
See https://github.com/geocoder-php/Geocoder/pull/1106#issuecomment-760456241

nihilsen:
> No, we do not "touch" the result returned by an API.

Got it. That seems reasonable.

jbelien (changes requested):
- We try to keep properties as defined by the API. Instead of `displayName`, could you rename it to `name`, `getName()`, `withName()` ?

- Could you create a new separated test (instead of including your test to `testGeocodeQuery()`) ?

Thanks.

nihilsen:
> Instead of displayName, could you rename it to name, getName(), withName() ?

You got it.

Tests separated.

jbelien (approved):

jbelien:
Thanks @nihilsen !

PR #1109, opened 2021-01-19
[Photon] Add support for geo position query parameters.

nihilsen:
The Photon search API supports the parameters `lat` and `lon` which serve to prioritize search results based on a given geo position.

This PR adds support for those parameters:

```PHP
$geocodeQuery = GeocodeQuery::create('Paris')
    ->withData('lat', 33.6625)
    ->withData('lon', -95.547778);
```

### DIFF
--- a/src/Provider/Photon/Photon.php
+++ b/src/Provider/Photon/Photon.php
@@ -77,6 +77,13 @@ final class Photon extends AbstractHttpProvider implements Provider
                 'lang' => $query->getLocale(),
             ]);
 
+        $lat = $query->getData('lat');
+        $lon = $query->getData('lon');
+
+        if (!is_null($lat) && !is_null($lon)) {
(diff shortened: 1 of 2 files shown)


src/Provider/Photon/Photon.php
```diff
@@ -77,6 +77,13 @@ public function geocodeQuery(GeocodeQuery $query): Collection
                 'lang' => $query->getLocale(),
             ]);
 
+        $lat = $query->getData('lat');
+        $lon = $query->getData('lon');
+
+        if (!is_null($lat) && !is_null($lon)) {
+            $url .= '&'.http_build_query(compact('lat', 'lon'));
```
jbelien, line 84:
```suggestion
            $url .= '&'.http_build_query([
                'lon' => $lon,
                'lat' => $lat
            ]);
```

jbelien (changes requested):

PR #1110, opened 2021-01-20
Expect a PSR-18 client instead of a PHP-HTTP client

### LINKED ISSUE #1076: Switch from PHP-HTTP to PSR-18
PHP-HTTP has been formalized into PSR-18 a while back. It would be really nice if we can switch to PSR-18 as this would allow us to use clients with 'native' PSR-18 support, but lacking PHP-HTTP support, such as Guzzle 7. Most (if not all) PHP-HTTP-clients also support PSR-18 so it should not be a big breaking change.

JaZo:
Resolves #1076 
Requires https://github.com/geocoder-php/provider-integration-tests/pull/18

N.B. It's a chicken and egg problem because in both pull requests the tests will fail because it needs changes from the other PR.


JaZo:
I can't change the integration tests of the providers because that requires a change in [provider-integration-tests](https://github.com/geocoder-php/provider-integration-tests), so I'll also create a PR for that and finish this PR after.

src/Http/Provider/AbstractHttpProvider.php
```diff
@@ -38,10 +38,10 @@ abstract class AbstractHttpProvider extends AbstractProvider
     private $messageFactory;
 
     /**
-     * @param HttpClient          $client
+     * @param ClientInterface     $client
      * @param MessageFactory|null $factory
      */
-    public function __construct(HttpClient $client, MessageFactory $factory = null)
+    public function __construct(ClientInterface $client, MessageFactory $factory = null)
```
Nyholm, line 44: This is okey because HttpClient extends ClientInterface.

src/Provider/AzureMaps/AzureMaps.php
```diff
@@ -69,7 +69,7 @@ class AzureMaps extends AbstractHttpProvider implements Provider
      * AzureMaps constructor.
      */
     public function __construct(
-        HttpClient $client,
+        ClientInterface $client,
```
Nyholm, line 72: Would you get a PHP error if you didn't change this?

src/Http/Provider/AbstractHttpProvider.php
```diff
@@ -106,9 +106,9 @@ protected function getParsedResponse(RequestInterface $request): string
     /**
      * Returns the HTTP adapter.
      *
-     * @return HttpClient
+     * @return ClientInterface
      */
-    protected function getHttpClient(): HttpClient
+    protected function getHttpClient(): ClientInterface
```
Nyholm, line 111: We are technically widening the return type. If someone extended this, they will get an error, right? Just as the `DummyProvider`.

Nyholm (commented):
Thank you. I think this looks good. 

I have a few small questions.

src/Provider/AzureMaps/AzureMaps.php
JaZo: PHP says it's fine, but it will make this PR useless as you are still tied to PHP-HTTP as HTTP-client.

src/Http/Provider/AbstractHttpProvider.php
JaZo: On PHP 7.3 they will get an error yes, but it is allowed as of PHP 7.4.

PavelVais:
Is there any progress of this PR ? It should be done to support guzzle7.

norkunas:
@JaZo are you willing to finish this? :)

JaZo:
Sorry for the long silence from my end. Yes, I still like to finish this! I'll go ahead and create a PR for [provider-integration-tests](https://github.com/geocoder-php/provider-integration-tests) so I can finish this PR.

JaZo:
I've opened https://github.com/geocoder-php/provider-integration-tests/pull/18 for the changes in the integration tests.

jbelien:
I've (finally) merged and released https://github.com/geocoder-php/provider-integration-tests/pull/18

Could you finalize this PR ?
Thanks! Sorry for the long delay.

src/Http/Provider/AbstractHttpProvider.php
jbelien: Support for PHP 7.3 has been dropped (since it has reached end of life) (see #1158)

jbelien:
@JaZo Could you let me know ALL the tags you created ?

You didn't create a tag with the correct semver number for a lot of the tags you created and I need to clean it!

PR #1118, opened 2021-05-05
[LocationIQ] Add autocomplete + countrycodes & tag filters

nassuf:
- Change locationIQ endpoint to the new one
- Add autocomplete endpoint
- Manage json to AddressCollection as autocomplete endpoint doest not return XML format
- Add countrycodes filter (filter within a bunch of countries + tag filter (type of address)

### DIFF
--- a/src/Provider/LocationIQ/LocationIQ.php
+++ b/src/Provider/LocationIQ/LocationIQ.php
@@ -32,7 +32,7 @@ final class LocationIQ extends AbstractHttpProvider implements Provider
     /**
      * @var string
      */
-    const BASE_API_URL = 'https://locationiq.org/v1';
+    const BASE_API_URL = 'https://api.locationiq.com/v1';
 
     /**
      * @var string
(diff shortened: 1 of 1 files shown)


src/Provider/LocationIQ/LocationIQ.php
```diff
@@ -58,15 +58,49 @@ public function __construct(HttpClient $client, string $apiKey)
...
+
+    /**
+     * {@inheritdoc}
+     */
+    private function autocompleteQuery(GeocodeQuery $query): Collection
     {
         $address = $query->getText();
+        $countrycodes = $query->getData('countrycodes');
+        $tag = $query->getData('tag');
+        $url = sprintf($this->getGeocodeAutocompleteEndpointUrl(), urlencode($address), $query->getLimit(), $countrycodes, $tag);
```
jbelien, line 77: What if `countrycodes` and/or `tag` is not set ?

src/Provider/LocationIQ/LocationIQ.php
```diff
@@ -108,6 +142,49 @@ public function reverseQuery(ReverseQuery $query): Collection
         return new AddressCollection([$this->xmlResultToArray($result, $addressParts)]);
     }
 
+    /**
+     * @param array $arrayResult
+     * @return Location
+     */
+    private function arrayResultToArray(array $arrayResult): Location
```
jbelien, line 149: Name of the function is confusing I think. I would name it something like `resultToLocation()`.

jbelien (changes requested):
Thanks a lot for your contribution @nassuf !

- Could you make sure your PR passes the StyleCI check ?
- Could you add a test for the autocomplete query ?
- Could you add a test for the `countrycodes` parameter ?
- Could you add a test for the `tag` parameter ?

If the search request also returns JSON, could you convert it so it's consistent with autocomplete request ?

Thanks!

PR #1122, opened 2021-05-12
LocationIQ - Adding state_code to AdminLevel

BrodyT:
Code for the state admin level was statically set to ''.


jbelien:
Thanks @BrodyT ,

Could you add a test for this ?

BrodyT:
@jbelien Sure, I'll look into adding that as soon as I can.

PR #1125, opened 2021-05-20
Replace empty() by more strict checks

### LINKED ISSUE #1124: GeocodeQuery with zero as text fails
When creating a geocode query with `'0'` as text, the value is handled as invalid character. `'0'` though can be a valid query, e.g. when querying for postal codes. The issue here is the usage of `empty()`.

```php

$query = \Geocoder\Query\GeocodeQuery::create('0');

```

jbelien:
The function `empty()` is a very loose comparison (see [manual](https://www.php.net/manual/en/function.empty)), it's recommended to use more strict comparison (also close #1124).

- Replace `empty($string)` for strings by `null !== $string && strlen($string) > 0`
- Replace `empty($array)` for arrays by `null !== $array && count($array) > 0`

---

I left `empty()` for all the `Bounds` check because I don't really see how to replace it by a more strict check but I think it should be updated as well.

@Nyholm What do you consider an "empty bound" ?

---

I left an `empty()` in [`AbstractArrayDumper`](https://github.com/geocoder-php/Geocoder/blob/c02b93a471a4e7043448b518183d449ceefa81b9/src/Common/Dumper/AbstractArrayDumper.php#L29-L31) because I think it makes sense to keep a loose check there.

---

@Nyholm Doing this check, I stumble upon something that seemed strange.
In the `LimitPlugin`, it checked [`$query->getLocale()`](https://github.com/geocoder-php/Geocoder/blob/1714d61578ea4881c8d481feae4e214e902095e4/src/Plugin/Plugin/LimitPlugin.php#L43-L45) but I think it should check [`$query->getLimit()`](https://github.com/geocoder-php/Geocoder/blob/c02b93a471a4e7043448b518183d449ceefa81b9/src/Plugin/Plugin/LimitPlugin.php#L43-L46) instead.
Let me know if it was correct and I missed something.

### DIFF
--- a/src/Common/Model/Address.php
+++ b/src/Common/Model/Address.php
@@ -235,12 +235,12 @@ class Address implements Location
 
         $adminLevels = [];
         foreach ($data['adminLevels'] as $adminLevel) {
-            if (empty($adminLevel['level'])) {
+            if (null === $adminLevel['level'] || $adminLevel['level'] === 0) {
                 continue;
             }
 
             $name = $adminLevel['name'] ?? $adminLevel['code'] ?? null;
-            if (empty($name)) {
+            if (null === $name || strlen($name) === 0) {
                 continue;
             }
 
--- a/src/Common/Model/Address.php
+++ b/src/Common/Model/Address.php
@@ -240,7 +240,7 @@ class Address implements Location
             }
 
             $name = $adminLevel['name'] ?? $adminLevel['code'] ?? null;
-            if (null === $name || 0 === strlen($name)) {
+            if (null === $name || '' === $name) {
                 continue;
             }
 
--- a/src/Common/Model/Address.php
+++ b/src/Common/Model/Address.php
@@ -235,12 +235,12 @@ class Address implements Location
 
         $adminLevels = [];
         foreach ($data['adminLevels'] as $adminLevel) {
-            if (null === $adminLevel['level'] || $adminLevel['level'] === 0) {
+            if (null === $adminLevel['level'] || 0 === $adminLevel['level']) {
                 continue;
             }
 
             $name = $adminLevel['name'] ?? $adminLevel['code'] ?? null;
-            if (null === $name || strlen($name) === 0) {
+            if (null === $name || 0 === strlen($name)) {
                 continue;
             }
 
--- a/src/Common/Model/AddressBuilder.php
+++ b/src/Common/Model/AddressBuilder.php
@@ -109,7 +109,7 @@ final class AddressBuilder
         }
 
         $country = null;
-        if (!empty($this->country) || !empty($this->countryCode)) {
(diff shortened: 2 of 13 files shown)


jbelien:
@Nyholm If you have a moment, could you have a look ? Thanks :)

Nyholm:
I'll try to check this tonight. Generally, I agree with removing the use of empty()

src/Common/Model/Address.php
```diff
@@ -235,12 +235,12 @@ public static function createFromArray(array $data)
...
         $adminLevels = [];
         foreach ($data['adminLevels'] as $adminLevel) {
-            if (empty($adminLevel['level'])) {
+            if (null === $adminLevel['level'] || 0 === $adminLevel['level']) {
                 continue;
             }
 
             $name = $adminLevel['name'] ?? $adminLevel['code'] ?? null;
-            if (empty($name)) {
+            if (null === $name || 0 === strlen($name)) {
```
Nyholm, line 243: This is quicker. 

```suggestion
            if (null === $name || '' ===$name) {
```

In symfony's code base we compare to the empty string instead of verifying the length.

src/Common/Model/AddressBuilder.php
```diff
@@ -109,7 +109,7 @@ public function build(string $class = Address::class): Address
         }
 
         $country = null;
-        if (!empty($this->country) || !empty($this->countryCode)) {
+        if ((null !== $this->country && strlen($this->country) > 0) || (null !== $this->countryCode && strlen($this->countryCode) > 0)) {
```
Nyholm, line 112: Same here. Compare with empty string.

src/Common/Model/AddressCollection.php
```diff
@@ -53,7 +53,7 @@ public function count()
      */
     public function first(): Location
     {
-        if (empty($this->locations)) {
+        if (0 === count($this->locations)) {
```
Nyholm, line 56:
```suggestion
        if ([] === $this->locations) {
```

Nyholm (commented):
I agree with all the changes. I'll just suggest to compare the variable with the empty string or empty array instead of checking the length.

src/Common/Model/Address.php
jbelien: Good to know! 👍 
See f00d9a1c6a93a7644816e3befea546d405af0bbe

src/Common/Model/AddressCollection.php
jbelien: See d118970ea7dc611448512e8a27a29e4f6fc7810e

src/Common/Model/AddressBuilder.php
jbelien: See 83f80c3490ccf11a8c9ce74cf7c2d7fd2bd5e7e0

Nyholm (approved):
Thank you

PR #1130, opened 2021-07-10
Check if geonames key exists

kornrunner:
Check if `geonames` key exists before accessing it (similar to #783)


jbelien (approved):

jbelien:
Thanks @kornrunner !
Update has been released: https://packagist.org/packages/geocoder-php/geonames-provider#4.3.1

PR #1131, opened 2021-07-14
[Nominatim] Set optional parameters of http_build_query

sebalis:
If arg_separator is not set, the default value comes from arg_separator.output which is not always set to the value we need here, which is '&'. On Ubuntu the default is '&amp;amp;'. There is no need to adhere to any system defaults here because we know exactly in which context the URLs we build are used: for a backend request that is executed soon in the same PHP handler. The same parameters are used in the MapQuest provider.

### DIFF
--- a/src/Provider/Nominatim/Nominatim.php
+++ b/src/Provider/Nominatim/Nominatim.php
@@ -98,7 +98,7 @@ final class Nominatim extends AbstractHttpProvider implements Provider
                 'addressdetails' => 1,
                 'extratags' => 1,
                 'limit' => $query->getLimit(),
-            ]);
+            ], '', '&', PHP_QUERY_RFC3986);
 
         $countrycodes = $query->getData('countrycodes');
         if (!is_null($countrycodes)) {
@@ -107,11 +107,11 @@ final class Nominatim extends AbstractHttpProvider implements Provider
 
                 $url .= '&'.http_build_query([
                     'countrycodes' => implode(',', $countrycodes),
-                ]);
+                ], '', '&', PHP_QUERY_RFC3986);
             } else {
                 $url .= '&'.http_build_query([
                     'countrycodes' => strtolower($countrycodes),
-                ]);
+                ], '', '&', PHP_QUERY_RFC3986);
             }
         }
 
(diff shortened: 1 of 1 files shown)


jbelien:
Hello @sebalis ,

Thank you for your contribution!

I must admit I'm a bit confused by this ...
I actually don't think it's actually an issue since all our tests are running on Ubuntu and I'm also running on Ubuntu without having any issue.
I just checked my PHP configuration (Ubuntu):
> ; The separator used in PHP generated URLs to separate arguments.
> ; PHP's default setting is "&".
> ; http://php.net/arg-separator.output
> ; Example:
> ;arg_separator.output = "\&amp;"

So if in your configuration `arg_separator.output` is set to `&amp;`, it's most likely a custom configuration and you're not using PHP's defaults.

PS: Nominatim tests are currently failing, I'll need to check why.

sebalis:
Hi @jbelien, thanks for responding!

I must admit that I was not sure myself why the constant is set to &amp;amp; on the Ubuntu instance where I encountered the error. I trired to find out which package installs ``/etc/php/*/*/php.ini`` and didn’t find one. Then I found that Ububtu ships suggested files for php.ini in the [php7.4-common](https://packages.ubuntu.com/focal/php7.4-common) package, which installs them under ``/usr/lib/php/7.4``/ as ``php.ini-{development,prroduction,production.cli}``. In all these files, arg_separator.output is set to "&amp;amp;". There seems to be an upstream source in [php7.4](https://packages.ubuntu.com/source/focal/php7.4): it contains an archive ``php7.4_7.4.3.orig.tar.xz`` which has two of these three files, both with the same value.

But my argument goes beyond this. The fact that calling http_build_query without setting these optional arguments introduces a flexibility that we don’t want. We know exactly which syntax we want to use for the URLs constructed in this file as we are going to use them for opening a server-to-server connection to Nominatim in the backend, straight after they are constructed. Getting those ini constants involved clearly serves another purpose – where the string is first written somewhere (say, as part of textual or HTML output) and then only used by some other process that works on this output (such as a browser). What is a welcome flexibility there is nothing but unwelcome uncertainty in this context. Let me again point to the MapQuest provider, which uses these same arguments (that is actually where I took them from).

sebalis:
I’ve only just realised that you seem to think using *&amp;amp;* is correct. In my situation it wasn’t – I needed *&amp;* to make my queries to nominatim.openstreetmap.org succeed. That’s why I used that value in the parameters I am proposing to add.

jbelien (approved):

jbelien:
Thanks for the reminder, it slipped my mind.

sebalis:
Thanks – I had wondered that maybe I should have created an issue together with the PR straight away :-)

jbelien:
No, no, it was not necessary.
I just forgot about it.

I'll fix the tests and create a new release ASAP.

PR #1133, opened 2021-07-19
Update Readme.md

oichou:


oichou:
Hello @jbelien 
I'm sorry for this misunderstanding, i closed  the last one to create a new one with a cleaner code with just one commit instead of having two

jbelien:
Oh okay, it was not necessary but not worries.
I'll check it ASAP.

jbelien (changes requested):
For consitency, could you write `new \Cache\Adapter\PHPArray\ArrayCachePool()` instead using `use` ?

Issue #1117, opened 2021-04-13
Mapbox - missing proximity query param


kmadenski:
Missing proximity query param: https://docs.mapbox.com/help/tutorials/local-search-geocoding-api/#the-proximity-parameter

PR #1134, opened 2021-08-20
[MapBox] Add proximity support

kmadenski:
geocoder-php/Geocoder#1117


jbelien:
Hello @kmadenski,

Thanks for your contribution!

Could you add a test for this new parameter ?

PR #1136, opened 2021-08-26
[Photon] Add country codes

nihilsen:
This simple PR adds country codes in the address results for the Photon provider.

Let me know if tests are necessary and I will look into getting those added.


jbelien (approved):
Thanks a lot @nihilsen 👍

PR #1139, opened 2021-09-26
[GoogleMapsPlaces] Add missing fields in model

comxd:
The current model for the GoogleMapsPlaces provider has some missing fields. This PR allows usage of all fields provided by Google Maps Places returned when mode=search. 

Added fields:
* business_status
* icon_background_color
* icon_mask_base_uri
* user_ratings_total
* reference


jbelien:
Nice, thanks @comxd !

Could you add some tests with those new fields ?

PR #1140, opened 2021-09-27
Specify iterable type

franmomu:
This will help static analysis tools to understand the iterable type of `Collection`


norkunas:
What's missing here to merge? :)

jbelien:
> What's missing here to merge? :)

I must admit I don't know that `@template-extends` and don't seem to find documentation about it.

Could someone give some documentation and context about what this PR actually does ?

norkunas:
In my understanding: `IteratorAggregate` can contain any element in the collection, but the geocoder stores the Location, so it extends and says to SA tools that this collection must store the `Location` implementing objects so we get the proper typehinting.

franmomu:
> In my understanding: `IteratorAggregate` can contain any element in the collection, but the geocoder stores the `Location`, so it extends and says to SA tools that this collection must store the `Location` implementing objects so we get the proper typehinting after calling `iterator_to_array`.

yes, `IteratorAggregate` [is defined as generic by Static Analysis tools](https://phpstan.org/blog/generics-in-php-using-phpdocs#type-safe-iterators-and-generators), about generics there is this article: https://phpstan.org/blog/generics-in-php-using-phpdocs#class-level-generics

So what this PR does is to tell SA tools (and PHPStorm) that `Collection` contains `Location` objects.

Here it's an example using this:

https://phpstan.org/r/ba1d6a3e-ba61-42db-9dde-a14169708c8b

so when you are iterating it for example, you know that each item is a `Location` object and can warn you if you are calling a method that does't exist in `Location` in this case.

jbelien (approved):
Thank you for the explanation!
I'll merge it and add PHPStan in the next few days.

jbelien:
Working on adding PHPStan in #1193 
I'll increase level bit by bit because there are 4783 errors for level 8. 😄

PR #1141, opened 2021-09-29
Allow psr/log 2.0 and 3.0

norkunas:


jbelien:
Thanks for your contribution @norkunas !

Could you explain the reason behind this upgrade ?

norkunas:
What to explain? :D I want that `composer outdated` would show empty list :stuck_out_tongue:

jbelien:
Sounds reasonable 😄 

Have you checked that it doesn't break anything ?
What happens if you use `psr/log:2.x` or `psr/log:3.x` ?

norkunas:
These versions just added typehints, I've checked the code and it complies contracts :)

jbelien (approved):

PR #1153, opened 2022-04-27
LocationIQ provider update from deprecated endpoint

Brew-Chris:
LocationIQ have sent out emails to customers informing them that LocationIQ are deprecating the following API endpoints from 1st June 2022: 
- https://us1.locationiq.org/v1
- https://eu1.locationiq.org/v1
- https://locationiq.org/v1
- https://locationiq.com/v1

The [LocationIQ API documentation](https://locationiq.com/docs-html/index.html#forward_usage) mention the two acceptable endpoints: 
- https://us1.locationiq.com/v1
- https://eu1.locationiq.com/v1

The LocationIQ Geocoder provider uses https://locationiq.org/v1 so it needs updating before 1st June 2022 or this provider will stop working.

This pull request is an update to the domains by way of a 3rd parameter to the constructor for "region". It will default to using "us1" but it will also accept "eu1"

The updated files in ./cached_responses are a result of running the tests with the new endpoint URL


src/Provider/LocationIQ/LocationIQ.php
```diff
@@ -43,13 +51,17 @@ final class LocationIQ extends AbstractHttpProvider implements Provider
...
      */
-    public function __construct(HttpClient $client, string $apiKey)
+    public function __construct(HttpClient $client, string $apiKey, string $region = null)
     {
         if (empty($apiKey)) {
             throw new InvalidCredentials('No API key provided.');
         }
 
         $this->apiKey = $apiKey;
+        if (!$region || !in_array($region, $this->regions)) {
```
jbelien, line 61:
```suggestion
        if (null === $region || true !== in_array($region, $this->regions, true)) {
```

I would also throw an error in the region is not valid (instead of using `us1`).

jbelien (changes requested):
Awesome! Thanks @Brew-Chris 👍

src/Provider/LocationIQ/LocationIQ.php
```diff
@@ -32,7 +33,15 @@ final class LocationIQ extends AbstractHttpProvider implements Provider
     /**
      * @var string
      */
-    const BASE_API_URL = 'https://locationiq.org/v1';
+    public $baseUrl = 'https://{region}.locationiq.com/v1';
```
jbelien, line 36: I don't think this needs to be `public` ; I would make it `private`.

jbelien (changes requested):
One last remark! 👍

src/Provider/LocationIQ/LocationIQ.php
Brew-Chris: I can make it protected if you want but the reason I made it public was because the constant it replaces was public and there may be users that are accessing it for some reason - which I didn't want to prevent.

jbelien:
```suggestion
    const BASE_API_URL = 'https://{region}.locationiq.com/v1';
```

You're right!
Removing the `BASE_API_URL` would create a breaking change anyway which is something we want to avoid.

I would keep the `BASE_API_URL` and use that constant to populate `$this->baseUrl` (and make that `baseUrl` property `protected` so there is no breaking change.

Constants are (still) `public` in our code because the visibility on class constants only came with PHP 7.1

Brew-Chris: I've reinstated the BASE_API_URL constant and used that to populate the now protected $baseUrl as requested.

Brew-Chris:
Sorry to bump but this pull request needs merging in and releasing within the next week or the LocationIQ provider will stop working on 1st June 2022.

jbelien:
Thanks for the reminder!
I'll take care of it end of this week.

jbelien (approved):

Issue #1155, opened 2022-05-14
Municipality not in list of 'localityFields'


uwbas:
I came across some reverse geocoding results where Nominatim.php didn't return a 'locality'. In my opinion, this latlng query should have 'Purmerend' as its locality, as it's the municipal boundary:

https://nominatim.openstreetmap.org/ui/reverse.html?lat=52.50741&lon=4.93279&zoom=18
https://nominatim.openstreetmap.org/ui/details.html?osmtype=W&osmid=6593763&class=highway

Upon debugging, I noticed that inside jsonResultToLocation(), on line 209, 'municipality' was not considered a 'locality'.

`$localityFields = ['city', 'town', 'village', 'hamlet'];`

Is this a matter of definition?

Adding 'municipality' fixes my issue:

`$localityFields = ['city', 'town', 'village', 'hamlet', 'municipality'];`

jbelien:
Hello @uwbas ,
That seems indeed to be a valid point!
Could you submit a PR with that change ?
Thanks.

uwbas:
Done: https://github.com/geocoder-php/Geocoder/pull/1156

PR #1156, opened 2022-05-28
[Nominatim] added 'municipality' as a 'locality'

uwbas:
Municipality was not in the array of 'locality' fields, see #1155.

Order matters, not sure what position in the array this should have.


jbelien:
Awesome! 👍 

Could you also add a test for it ?
Don't forget to also commit the cached response.

PaulPatat:
This is not consistent with the GoogleMaps provider at least for Dutch addresses. The town of Zaandam is in the municipality of Zaanstad. Google maps returns Zaandam as locality and Zaanstad as administrative area level 2. This would return Zaanstad as locality and nothing as administrative area level 2 (we don't have counties).

PR #1168, opened 2022-11-10
Include the postalcode in the reverse query as well in TomTom provider

Spriz:
I'd guess this line was left out by accident - but never the less it would be nice to have it here as well :-)


Spriz:
I just copied the line from https://github.com/Spriz/Geocoder/blob/patch-1/src/Provider/TomTom/TomTom.php#L102 btw

jbelien:
Thanks a lot @Spriz 👍 
Could you add the postal code to the test as well ? Thanks!

Spriz (commented):
Check for postal code being not-null in tests as well

Spriz:
@jbelien Done! :+1:

jbelien:
Thanks @Spriz !
A new release has been published: https://packagist.org/packages/geocoder-php/tomtom-provider#4.4.0

PR #1178, opened 2023-02-25
[Nominatim] Add support for neighbourhood

### LINKED ISSUE #1167: [Nominatim] Neighbourhood field
Some addresses have a "neighbourhood" field, for example : 

https://nominatim.openstreetmap.org/reverse?format=xml&lat=35.685939&lon=139.811695&zoom=18&addressdetails=1

In jsonResultToLocation(), this field is not processed, so we can't get that value through the reverseQuery() method.

jbelien:
Add support for `address.neighbourhood` property

Close #1167


Nyholm (approved):
Thank you

PR #1188, opened 2023-07-03
[Photon] Add state, county & district properties

ybert:
Add state, county and district properties from Photon API.


jbelien (approved):
Thanks a lot 👍 
I'll merge it and create a new release in the next few days!

PR #1195, opened 2023-07-20
[Photon] Reverse Geocoding with query filters

ybert:
It can be usefull to search precise data from osm when we do reverse geocoding.

For example it can help to search only osm places (locality, city) from a location (latitude, longitude)


jbelien (changes requested):
Hello @ybert ,

Thanks a lot for your contribution.
I've looked a bit for documentation about `query_string_filter` but couldn't find anything. After a bit more research, I've found https://github.com/komoot/photon/issues/703#issuecomment-1461729947.

I see you've commented this issue but it seems that that filter was not really intended to be implemented and might be removed.
> In theory we have the query_string_filter parameter for reverse which got sneaked in in https://github.com/komoot/photon/pull/254. It should be able to do arbitrary filtering including the one you want. However, this was never documented and I very much prefer to let it die quietly and implement the tag filters from search instead.

I would prefer not to implement this feature in our provider (because we will have to remove it at some point which will create a breaking change) and instead implement `osm_tag` for the search (and the reverse geocoding once implemented).

ybert:
Ok, I undersand your position.
Actually it is the only way to use reverse geocoding with photon and have specific result types. 

I will see if I have time to propose a PR for osm_tag on the Photon repository.

ybert:
[My PR](https://github.com/komoot/photon/pull/742 ) has just been merged in Photon master branch.

Can I implement geocoding and reverse with osm tags here or should I wait a tagged version in photon ?

jbelien:
> https://github.com/komoot/photon/pull/742 has just been merged in Photon master branch.

Awesome work! 👍 

> Can I implement geocoding and reverse with osm tags here or should I wait a tagged version in photon ?

Of course, you can definitely already update this PR (or close this PR and open a new one if you prefer).

ybert:
@jbelien Any chance to merge this PR ?

src/Provider/Photon/Tests/PhotonTest.php
```diff
@@ -88,21 +89,54 @@ public function testGeocodeQueryWithNamedResult(): void
...
+            $this->assertEquals('tourism', $result->getOSMTag()->key);
+            $this->assertEquals('museum', $result->getOSMTag()->value);
+        }
+    }
+
+    public function testGeocodeQueryWithMultipleOsmTagFilter(): void
+    {
+        $provider = Photon::withKomootServer($this->getHttpClient());
+        $query = GeocodeQuery::create('Paris')
+            ->withData('osm_tag', ['tourism', ':!museum'])
```
jbelien, line 118: Not sure passing an array here makes sense.
I see you added the `:` in the value ; I think it would make more sense to only allow string for the value (so `tourism:museum` or `tourism:!museum`).

Am I missing something ?

src/Provider/Photon/Photon.php
```diff
@@ -158,6 +167,25 @@ public function getName(): string
         return 'photon';
     }
 
+    /**
+     * @param string|array<int, string>|null $filters
+     */
+    private function buildOsmTagFilterQuery($filters): string
```
jbelien, line 173:
```suggestion
    private function buildOsmTagFilterQuery(?string $filters): string
```

See comment in `PhotonTest.php`.

jbelien (changes requested):

src/Provider/Photon/Tests/PhotonTest.php
ybert: In this particular example you're right.
But what if I want to search for museums and art galleries ?
```php
        $provider = Photon::withKomootServer($this->getHttpClient());
        $query = GeocodeQuery::create('Paris')
            ->withData('osm_tag', ['tourism:museum', 'tourism:gallery'])
```

jbelien: Good point! Could you update your test accordingly and make just that multiple filters logic works as expected ? Thanks.

ybert: I just updated my tests.
The multiple osm tag logic already works as expected.

jbelien (approved):

ybert:
@jbelien Can we make a new release for the Photon provider to make these changes available ?

jbelien:
> @jbelien Can we make a new release for the Photon provider to make these changes available ?

Sure! But you need to merge your PR first! 😉

ybert:
I think I don't have write access to merge this PR.

ybert:
@jbelien How can I merge this PR ? I think I don't have write access.

ybert:
@jbelien Let me know if there is anything I can do to get it merged ?

jbelien:
@ybert My apologies, you indeed don't have the write to merge that PR. Taking care of it right now.

PR #1197, opened 2023-07-21
Drop support for PHP 7.4

jbelien:
PHP 7.4 has reach end of life in November 2022 (see https://www.php.net/supported-versions.php)

> **Note**  
> Deprecate Geoip provider since geoip extension doesn't seem to be available with PHP 8 (README updated to suggest to switch to GeoIP2 provider)


jbelien:
Other than removing support for a version of PHP that has reached end of life is also to be able to use [`mixed`](https://www.php.net/manual/language.types.mixed.php) in #1194

PR #1201, opened 2023-10-10
Add IP2LocationIo provider

ip2location:
Added service from https://www.ip2location.io as a new provider.


jbelien:
Hello @ip2location,

We already have a IP2Location provider: https://github.com/geocoder-php/Geocoder/tree/master/src/Provider/IP2Location

What's the difference between IP2Location.com and IP2Location.io ?

ip2location:
Hello @jbelien,

The IP2Location is using an old API interface which we have deprecated. However, we are still supporting it for all existing users.

This PR is for IP2Location.io API which he launched in 2021. The new API is faster and offer free queries.

ip2location:
Hi @jbelien, 

Do you need more information from us? Feel free to ask.

m-cameron:
To ensure legacy support, we will maintain the existing implementation at api.ip2location.com for users still on the old interface. The new interface, IP2Location.io, is hosted on a separate domain. 

Therefore, creating a new provider, IP2LocationIO, will prevent confusion while allowing us to support both legacy and new API users.

ip2location:
Hi @jbelien

To ensure legacy support, we will maintain the existing implementation at api.ip2location.com for users still on the old interface. The new interface, IP2Location.io, is hosted on a separate domain. 

Therefore, creating a new provider, IP2LocationIO, will prevent confusion while allowing us to support both legacy and new API users.

jbelien:
Okay, I get it ; but is the stack and functionalities behind IP2Location and IP2LocationIo completely different ?
With a new major version you can still easily maintain both in the same provider (with 2 separate branches).

In any case, could you add a paragraph in **both** READMEs for IP2Location and IP2LocationIo that explain the difference ? Thanks.

PR #1203, opened 2023-10-24
Pelias/custom address fields

mrbig:
Pelias is an open source modular geocoder. It support several data sources and tries to find the best match for each query. Unfortunately a lot of this information is hidden in PHP Geocoder. This PR will expose the most important fields through an extended PeliasAddress class.

The solution is similar to the GoogleMaps provider's GoogleAddress, I have also included the with... methods although currently they are not in use.

The change also includes a minor refactor in the provider that leads to better extendability if one would want to override how the results are parsed into addresses.

### DIFF
--- a/src/Provider/Pelias/Model/PeliasAddress.php
+++ b/src/Provider/Pelias/Model/PeliasAddress.php
@@ -1,0 +1,175 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Geocoder package.
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @license    MIT License
+ */
+
+namespace Geocoder\Provider\Pelias\Model;
+
+use Geocoder\Model\Address;
+
+class PeliasAddress extends Address
+{
+    /**
+     * The pelias layer returned
+     * @var string|null
+     */
+    private $layer;
+
+    /**
+     * Confidence score from pelias
+     * @var float|null
+     */
+    private $confidence;
+
+    /**
+     * Match type from pelias
+     * @var string|null
+     */
+    private $matchType;
+
+    /**
+     * Data source from pelias
+     * @var string|null
+     */
+    private $source;
+
+    /**
+     * Accuracy from pelias
+     * @var string|null
+     */
+    private $accuracy;
+
+    public static function createFromArray(array $data)
+    {
+        $address = parent::createFromArray($data);
+        $address->layer = $data['layer'] ?? null;
+        $address->confidence = $data['confidence'] ?? null;
+        $address->matchType = $data['match_type'] ?? null;
+        $address->source = $data['source'] ?? null;
+        $address->accuracy = $data['accuracy'] ?? null;
+        return $address;
+    }
+
+    /**
+     * Get the pelias layer returned
+     *
+     * @return  string|null
+     */ 
+    public function getLayer()
+    {
+        return $this->layer;
+    }
+
+    /**
+     * Get confidence score from pelias
+     *
+     * @return  float|null
+     */ 
+    public function getConfidence()
+    {
+        return $this->confidence;
+    }
+
+    /**
+     * Get match type from pelias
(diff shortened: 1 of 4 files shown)


src/Provider/Pelias/Model/PeliasAddress.php
```diff
@@ -0,0 +1,175 @@
...
+
+use Geocoder\Model\Address;
+
+class PeliasAddress extends Address
+{
+    /**
+     * The pelias layer returned
+     * @var string|null
+     */
+    private $layer;
```
norkunas, line 23: Why not defining real typehints?

src/Provider/Pelias/Pelias.php
```diff
@@ -132,48 +133,62 @@ protected function executeQuery(string $url): AddressCollection
...
+                'east' => $location['bbox'][0],
+            ];
+        } else {
+            $bounds = [
+                'south' => null,
+                'west' => null,
+                'north' => null,
+                'east' => null,
+            ];
+        }
```
norkunas, line 163:
```suggestion
        $bounds = [
            'south' => $location['bbox'][3] ?? null,
            'west' => $location['bbox'][2] ?? null,
            'north' => $location['bbox'][1] ?? null,
            'east' => $location['bbox'][0] ?? null,
        ];
```

src/Provider/Pelias/Pelias.php
```diff
@@ -132,48 +133,62 @@ protected function executeQuery(string $url): AddressCollection
...
+            }
         }
 
-        return new AddressCollection($results);
+        return PeliasAddress::createFromArray([
+            'providedBy' => $this->getName(),
+            'latitude' => $location['geometry']['coordinates'][1],
+            'longitude' => $location['geometry']['coordinates'][0],
+            'bounds' => $bounds,
+            'streetNumber' => isset($props['housenumber']) ? $props['housenumber'] : null,
```
norkunas, line 178: null coalescing could be used

jbelien (changes requested):
Thanks @mrbig !

Could you add documentation for the properties you've added. Are those properties **always** return by any Pelias instance ?

Please also have a look at the comments made by @norkunas.

Thanks!

src/Provider/Pelias/Model/PeliasAddress.php
mrbig: I have added real type hints.

src/Provider/Pelias/Pelias.php
mrbig: My goal was to change the original Provider as minimally as possible.
However this is a nice touch to make the code more readable, so I have included your suggestion.

src/Provider/Pelias/Pelias.php
mrbig: I have replaced the code wherever possible

mrbig:
@jbelien I have implemented the changes suggested by @norkunas.

Could you point me to the documentation that I should update? I've included documentation in the code itself, I realize that might not be sufficient.

The fields I have added are always returned to my knowledge. They describe what data source was used, how accurate that data source is considered, could it find the exact location or it did some kind of interpolation.

Our use case:
We are geocoding a large dataset using opensource data. Unfortunately some regions aren't exact in these databases at the moment. So we want to store with the coordinates the accuracy at the time of geocoding, so that after a map update we can go reiterate these addresses and run another geocoding to refine them. But if the result of geocoding change without the accuracy information we even can not decide if the new values are better or worst.

jbelien:
> Could you point me to the documentation that I should update? I've included documentation in the code itself, I realize that might not be sufficient.

In `src/Provider/Pelias/Model/PeliasAddress.php`, could you add the link to the documentation ?

For instance:
```php
/**
 * @link https://github.com/pelias/documentation/blob/master/result_quality.md#confidence Confidence score
 * @var float|null
 */
private ?float $confidence = null;
```

mrbig:
Hi @jbelien do you need any other change on this PR?

PR #1209, opened 2023-12-21
Prepare release of 5.0.0 for willdurand/geocoder

Nyholm:
Is there anything else we should add here? 

https://github.com/geocoder-php/php-common/compare/4.6.0...master

### DIFF
--- a/.github/workflows/component.yml
+++ b/.github/workflows/component.yml
@@ -14,8 +14,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['8.0', '8.1', '8.2', '8.3']
+        php-version: ['8.2', '8.3']
         component: [Common, Http, Plugin]
+        include:
+          - php-version: '8.0'
+            component: Http
+          - php-version: '8.0'
+            component: Plugin
+          - php-version: '8.1'
+            component: Http
+          - php-version: '8.1'
+            component: Plugin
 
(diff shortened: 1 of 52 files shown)


jbelien:
I'll investigate the checks that are failing on PHP 8.2 but the tests for BingMaps seems to fail only on PHP 8.3.
> 1) Geocoder\Provider\BingMaps\Tests\BingMapsTest::testGeocodeWithRealAddressReturnsMultipleResults
unserialize(): Extra data starting at offset 3984 of 3985 bytes

Chris53897:
@jbelien I guess it was wrong all the time and just popped up with PHP 8.3.
https://php.watch/versions/8.3/unserialize-E-WARNING

I did create a PR (see above) to solve this.
I am not sure if this is the best way. But i could not figure out where the LINE Carriage (i guess) at the File-End is created in the first place.

jbelien:
✅ We're back on track: #1220

Nyholm:
I've rebased the PR. 

@jbelien please merge if you think we are ready.

.github/workflows/component.yml
```diff
@@ -14,8 +14,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['8.0', '8.1', '8.2', '8.3']
+        php-version: ['8.2', '8.3']
```
Chris53897, line 17: Could you please add '8.4' here and in the other places?

jbelien (approved):

PR #1211, opened 2023-12-21
Revert subtree action back to checkout v3

Nyholm:
This PR https://github.com/geocoder-php/Geocoder/pull/1200 broke our subtree split. I reverse it now to get it working.

### DIFF
--- a/.github/workflows/subtree.yml
+++ b/.github/workflows/subtree.yml
@@ -19,7 +19,7 @@ jobs:
     name: Subtree for Common
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


jbelien:
Sorry about that, the issue was not caused by #1200 (so I've reverted it in 983e585224a296f701972570a1b637f8655fbbe7) but by my personal token that has expired.
I've rotated the token (will expire on 1st December 2024) and updated `SUBTREE_GITHUB_TOKEN` secret in Geocoder PHP organization secrets.

Nyholm:
Wohoo. Awesome. 

Yeah. I noticed this too. Thank you for the fix.

PR #1214, opened 2024-01-31
Add Faker provider

ker0x:
This PR introduce a faker provider using [FakePHP](https://fakerphp.github.io/). It's main purpose is to provide fake data during tests.


jbelien:
Hello @ker0x, welcome to the project.
I'm not exactly sure I understand the goal of that new provider, could you give us some use cases ?
Thanks.

ker0x:
@jbelien yes sure.

In my case, I have a listener on some of my entities that geocodes them each time they are created or updated. During my tests, I perform a lot of operations on these entities, and using a real provider slows down the tests because of HTTP calls. I don't particularly need the geocoded data to be accurate during tests, so having a provider with fake data saves me from making these HTTP calls (and it saves requests with some providers that have free request limits per month).

Chris53897:
@ker0x Could you please revert the changes to other files beside from your new provider, and rebase against master.
This wil make the ci green.

composer.json
```diff
@@ -30,6 +30,7 @@
         "cache/array-adapter": "^1.0",
         "cache/simple-cache-bridge": "^1.0",
         "cache/void-adapter": "^1.0",
+        "fakerphp/faker": "^1.23",
```
jbelien, line 33: Please revert. That should be part of the `composer.json` of your provider, not our main `composer.json`.

jbelien (changes requested):

composer.json
ker0x: @jbelien It's done!

PR #1223, opened 2024-03-02
feat: remove unnecessary provider interface

Chris53897:
This is already implemented via the parent classes https://github.com/geocoder-php/Geocoder/blob/master/src/Common/Provider/AbstractProvider.php


jbelien:
Indeed not needed but I believe it makes it more clear.

PR #1225, opened 2024-05-16
Utf8 encode

kornrunner:
Trying to avoid `utf8_encode` deprecation warning on PHP 8.2+

It will try to do the same using `mbstring` or `iconv`  functions depending on available extension, and eventually fallback to utf8_encode (not to break BC?).

It's a bit of duplicated code - but I didn't think this specific case (used in two providers only) would warrant moving it to parent class.

### DIFF
--- a/src/Provider/IP2LocationBinary/IP2LocationBinary.php
+++ b/src/Provider/IP2LocationBinary/IP2LocationBinary.php
@@ -81,9 +81,9 @@ final class IP2LocationBinary extends AbstractProvider implements Provider
             Address::createFromArray([
                 'providedBy' => $this->getName(),
                 'countryCode' => $records['countryCode'],
-                'country' => null === $records['countryName'] ? null : utf8_encode($records['countryName']),
+                'country' => null === $records['countryName'] ? null : mb_convert_encoding($records['countryName'], 'UTF-8', 'ISO-8859-1'),
                 'adminLevels' => $adminLevels,
-                'locality' => null === $records['cityName'] ? null : utf8_encode($records['cityName']),
+                'locality' => null === $records['cityName'] ? null : mb_convert_encoding($records['cityName'], 'UTF-8', 'ISO-8859-1'),
                 'latitude' => $records['latitude'],
                 'longitude' => $records['longitude'],
                 'postalCode' => $records['zipCode'],
--- a/src/Provider/IP2LocationBinary/composer.json
+++ b/src/Provider/IP2LocationBinary/composer.json
@@ -14,6 +14,7 @@
     "require": {
         "php": "^8.0",
         "ip2location/ip2location-php": "^8.1.1",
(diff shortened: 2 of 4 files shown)


jbelien:
> [!NOTE]
> Reference: https://www.php.net/manual/en/function.utf8-encode.php#refsect1-function.utf8-encode-notes

Thanks a lot ; I indeed miss `utf8_encode()` was deprecated!
To be honest, I think we could fully rely on `mbstring` only and replace `utf8_encode()` by `mb_convert_encoding()`. We "just" need to add `mbstring` as required extension in the `composer.json` file for those 2 providers (and create a new major version).

kornrunner:
Sure thing, that works as well (and is simpler). I wasn't sure what's the policy on BC changes.

PR is updated as suggested. Thank you!

jbelien:
Thanks @kornrunner 👍 
Don't forget to make sure all checks are passing.

kornrunner:
Both failing checks (phpstan and php cs fixer) seem to be regarding code that hasn't been affected in this PR (GogoleMaps and GeoIP2). 

I can submit another PR regarding those - as I don't think they address current issue?

norkunas:
> We "just" need to add `mbstring` as required extension in the `composer.json` file for those 2 providers (and create a new major version).

Another option would be to require [symfony/polyfill-mbstring](https://packagist.org/packages/symfony/polyfill-mbstring) to avoid breaks for apps that doesn't have mbstring installed

kornrunner:
Changed requirement from `mbstring` extension to [symfony/polyfill-mbstring](https://packagist.org/packages/symfony/polyfill-mbstring)

jbelien:
As soon you've merged #1226 and rebase this PR, I'll approve it! Thanks a lot! :+1:

kornrunner:
Sure thing, just - I have no privileges to merge it. Thanks!

jbelien:
Ah that's strange ... I've just done it 👌

kornrunner:
It's rebased, thank you!

jbelien (approved):

kornrunner:
Weirdly - it seems this change hasn't propagated to https://github.com/geocoder-php/ip2location-binary-provider ?

jbelien:
Indeed! I'll investigate.

jbelien:
IP2Location is indeed not in the list : https://github.com/geocoder-php/Geocoder/actions/runs/9142343555

There is probably a reason. I'll check and add it if it's an oversight.

jbelien:
🔖 IP2LocationBinary version 1.3.1: https://packagist.org/packages/geocoder-php/ip2location-binary-provider#1.3.1
🔖 MaxMindBinary version 4.3.1: https://packagist.org/packages/geocoder-php/maxmind-binary-provider#4.3.1

kornrunner:
Thank you!

PR #1229, opened 2024-06-27
Update Here Geocode and Reverse Geocode Provider to GS7 (v7)

SCatrinh:
Updated Here.com provider to Here GS7.

We are currently using here-provider 0.6.0 that is using Here v6.2 and are updating to use Here's GS7 endpoints.

Note: with the change to GS7 the shape of the responses from here has completely changed
Update was made using the following Here.com Documentation.

Edit: Also includes an update to php common to allow for providers to use alternative query methods to query strings IE Here.com's qualified queries.

Here Migration Docs
https://www.here.com/docs/bundle/geocoding-and-search-api-migration-guide/page/migration-geocoder/README.html

Geocode Endpoint Docs
https://www.here.com/docs/bundle/geocoding-and-search-api-developer-guide/page/topics/endpoint-geocode-brief.html

Reverse Geocode Endpoint Docs
https://www.here.com/docs/bundle/geocoding-and-search-api-developer-guide/page/topics/endpoint-reverse-geocode-brief.html

Here API Reference
https://www.here.com/docs/bundle/geocoding-and-search-api-v7-api-reference/page/index.html


SCatrinh:
Closing, determining a better solution. Thank-you for your time and patience.

PR #1231, opened 2024-07-03
Gs7 updateUpdate Here Geocode and Reverse Geocode Provider to GS7 (v7)

### LINKED ISSUE #1230: Update Here Geocode and Reverse Geocode Provider to GS7 (v7)
Here.com has released their GS7 (v7) API.

We are currently using here-provider 0.6.0 that is using v6.2 endpoints. We are updating to use Here's GS7 endpoints per a conversation with our Here.com rep. Our rep let us know that Here.com intends to drop support for our current endpoints (v6.2) and that we should switch to the new GS7 endpoints.

Note: with the change to GS7 the shape of the responses from here has completely changed
Update was made using the following Here.com Documentation. 

Here Migration Docs
https://www.here.com/docs/bundle/geocoding-and-search-api-migration-guide/page/migration-geocoder/README.html

Geo-code Endpoint Docs
https://www.here.com/docs/bundle/geocoding-and-search-api-developer-guide/page/topics/endpoint-geocode-brief.html

Reverse Geo-code Endpoint Docs
https://www.here.com/docs/bundle/geocoding-and-search-api-developer-guide/page/topics/endpoint-reverse-geocode-brief.html

Here API Reference
https://www.here.com/docs/bundle/geocoding-and-search-api-v7-api-reference/page/index.html

SCatrinh:
Updates Here provider to support GS7 under the 0.8.0 version tag

Closes #1230


patrickghidossi:
Synced with master

jbelien:
Hello @patrickghidossi ,

Thanks for your contribution.

I had a first (quick) look to the changes made.
This looks more like a breaking change (and so a new major version) than just an update (and minor version).

Looking at some [documentation](https://demo.support.here.com/hlsm/here_geocoding_and_search_v7), I've found this:
> GS7 introduces a new way to authenticate the application. Instead of using the Appid/Appcode combination, you must use the API Key or OAuth 2.0 token authentication methods for improved security. App code credentials are now on the verge of deprecation as they are the least secure of the authentication credential types.
I still see some reference to AppId in your code, is it stil supported?

I see that you changed some tests, the goal of those tests is to make sure we don't introduce breaking changes ; please keep the tests as they are, only adapt what needs to be adapted. If there are new features in v7, feel free to add **extra** tests.

Before I'll review the code, could you already do the following:
- Prepare version `1.0.0` instead of `0.8.0`
- Rebase to our `master` branch
- Make sure all PHPStan checks are passing
- Keep tests as they are

Thanks!

PR #1236, opened 2024-11-14
Feature/fix php cs

Chris53897:


ruudk:
@Chris53897 Please rebase this against main, so we can merge this.

ruudk:
@jbelien Can we merge this as is? I think that solves the problem on master, as that one is now broken. Thanks 🙏

jbelien (approved):

PR #1242, opened 2025-01-30
 Add support for displaying municipality in AzureMaps autocomplete results

kryber:
We noticed that the autocomplete functionality in the AzureMaps provider was missing the ability to display the municipality (city). This was causing incomplete location suggestions, as cities were not included in the results.

To fix this, we added the following line to the formatGeocodeResponse function:

`$builder->setLocality($result->address->municipality ?? null);`

Now, when searching for an address like Via Giuseppe Garibaldi 62, IT, Italy, the autocomplete properly returns results that include city names. This improves the accuracy and usability of the geocoding provider.

Let us know if any adjustments are needed! 🚀

image


jbelien:
All good 👍 Thanks for your contribution!

Could you just add a test about this change?
I'll then approve and merge it. Thanks.

kryber:
I've added a test to verify that the municipality (city) is properly included in the geocode response. Let me know if anything more is needed. Thank you for your fantastic job!

jbelien:
Could you make sure all failing checks are passing?

PHP CS Fixer is "just" a linting/formatting issue.
To fix the other tests, you'll have to commit the cached response of your new test.

Thanks!

jbelien (approved):

PR #1245, opened 2025-02-10
[Photon] Improve locality for city layer & Add layer & radius query params

ybert:
The idea here is to get the city in the PhotonAddress Model when the object returned by Photon is of type "city".
For example : [https://photon.komoot.io/reverse?lon=13.38890&lat=52.51644&lang=default&limit=1&layer=city&radius=10](https://photon.komoot.io/reverse?lon=13.38890&lat=52.51644&lang=default&limit=1&layer=city&radius=10)

In this example, we don't have the field city returned by Photon because it is the city element itself (so in this case the city/locality name is in the `name` field).

I also added :
- support for `radius` and `layer` query params as documented here [https://github.com/komoot/photon](https://github.com/komoot/photon)
- wrapping `type` in PhotonAddress model

### DIFF
--- a/src/Provider/Photon/Model/PhotonAddress.php
+++ b/src/Provider/Photon/Model/PhotonAddress.php
@@ -62,7 +62,7 @@ final class PhotonAddress extends Address
     public function getLocality(): ?string
     {
         $locality = parent::getLocality();
-        if ($locality === null && $this->type === 'city' && $this->name !== null) {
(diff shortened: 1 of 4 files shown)


jbelien:
Thanks for you contribution! Looks good 👍 

Could you rebase your PR to our `master` branch so I can approve and merge it? Thanks.

jbelien (approved):

PR #1247, opened 2025-02-18
[Photon] Support bbox in geocode query

tobiasbaehr:
I hope there is no hidden method or way I could do this without the changes.


src/Provider/Photon/Photon.php
```diff
@@ -190,6 +194,20 @@ private function buildOsmTagFilterQuery($filters): string
...
+        if (null === $query->getBounds()) {
+            return null;
+        }
+
+        return '&bbox='.sprintf('%s,%s,%s,%s',
+            $query->getBounds()->getWest(),
+            $query->getBounds()->getSouth(),
+            $query->getBounds()->getEast(),
+            $query->getBounds()->getNorth()
+        );
```
jbelien, line 208: I would use `sprintf('%.6f,%.6f,%.6f,%.6f', <bounds>)` or `implode(',', <bounds>)`

src/Provider/Photon/Photon.php
```diff
@@ -76,6 +76,10 @@ public function geocodeQuery(GeocodeQuery $query): Collection
         if (!empty($osmTagFilters)) {
             $url .= $osmTagFilters;
         }
+        $bboxQueryString = $this->buildBboxFilterQuery($query);
+        if ($bboxQueryString) {
```
jbelien, line 80:
```suggestion
        if (!is_null($bboxQueryString)) {
```

jbelien (changes requested):

jbelien (approved):

PR #1248, opened 2025-03-15
feat: [Photon] support lat/lon in geocode query

fbuchlak:
add lat/lon parameters support for photon (search with priority to a geo position)


src/Provider/Photon/Tests/PhotonTest.php
```diff
@@ -135,6 +135,25 @@ public function testGeocodeQueryWithMultipleOsmTagFilter(): void
         $this->assertGreaterThan(0, $countGalleries);
     }
 
+    public function testGeocodeQueryWithLatLon(): void
+    {
+        $provider = Photon::withKomootServer($this->getHttpClient());
+
+        $query = GeocodeQuery::create('Paris')->withLimit(1);
+        $results = $provider->geocodeQuery($query);
+        $this->assertInstanceOf('Geocoder\Model\AddressCollection', $results);
```
jbelien, line 144:
```suggestion
        $this->assertInstanceOf(AddressCollection::class, $results);
```

Don't forget to import `AddressCollection`.

src/Provider/Photon/Tests/PhotonTest.php
```diff
@@ -135,6 +135,25 @@ public function testGeocodeQueryWithMultipleOsmTagFilter(): void
...
+        $this->assertCount(1, $results);
+        $this->assertEquals('France', $results->first()->getCountry());
+
+        $query = $query
+            ->withData('lat', 33.661426)
+            ->withData('lon', -95.556321);
+        $results = $provider->geocodeQuery($query);
+        $this->assertInstanceOf('Geocoder\Model\AddressCollection', $results);
```
jbelien, line 152:
```suggestion
        $this->assertInstanceOf(AddressCollection::class, $results);
```
Don't forget to import `AddressCollection`.

jbelien (changes requested):

src/Provider/Photon/Tests/PhotonTest.php
fbuchlak: There are multiple assertions like this in that test file. Is it ok to update them all in this PR, or should I change them only in new test?

jbelien: Only focus on the ones in the Photon provider to keep the PR clear and simple!
If you have the energy and time to do it everywhere, that would be awesome but I would do it in a separate PR. Thanks.

fbuchlak: Resolved in this test. Leaving other cases for another PR.

jbelien (approved):

PR #1251, opened 2025-03-21
[GraphHopper] Add provider parameter

bisleadity:
Graphhopper geocoding now supports different providers. Hereby the option is given to use those providers.

As we do need to use on of those providers and there was no possibility given so far I added it.


src/Provider/GraphHopper/GraphHopper.php
```diff
@@ -74,6 +74,11 @@ public function geocodeQuery(GeocodeQuery $query): Collection
             $url .= sprintf('&bbox=%f,%f,%f,%f', $bounds->getWest(), $bounds->getSouth(), $bounds->getEast(), $bounds->getNorth());
         }
 
+        $provider = $query->getData('provider');
+        if (is_string($provider) && '' !== $provider) {
+            $url .= sprintf('&provider=%s', urlencode($provider));
+        }
```
jbelien, line 80: Maybe worth creating an [`enum`](https://www.php.net/manual/en/language.types.enumerations.php) with the list of available providers?

jbelien: Does all providers return the necessary result?

jbelien (commented):
This is a feature still in development.
> The provider parameter is currently under development and can fall back to default at any time.
> *-- <https://docs.graphhopper.com/openapi/geocoding/getgeocode#geocoding/getgeocode/t=request&in=query&path=provider>*

src/Provider/GraphHopper/GraphHopper.php
bisleadity: Enum would be no problem, but when the list expands another PR has to be made.

What do you mean by 'Does all providers return the necessary result?'. We do need one specific provider that works best for our case

jbelien: > What do you mean by 'Does all providers return the necessary result?'. We do need one specific provider that works best for our case

Is the response from GraphHopper different if we select a provider other than 'default'? Just to know if we need to implement specific logic if we choose another provider than the default one.

bisleadity: I tested a little and the response seems to be quiety diffrent depending on the address searched. Latitude and longitude are always present. But there are parameters , which are set depending on provider but as well the search string. So I'm not sure if its really useful to add them. However, the main address attributes such as city, country, street, housenumber or postcode seem to be the same for all providers.

One question: I was already wondering why the classes are made final if not so a user could extend e.g. the GraphHopper provider and Address Model to add properties if needed.

jbelien:
Could you maybe still add a real test?
Thanks!

jbelien:
@bisleadity Sorry, I mean a test that makes a geocoding request on GraphHopper with the new `provider` parameter.

Thanks!

PR #1252, opened 2025-03-23
style: convert string class names to constants

fbuchlak:
Based on discussion in #1248 this PR converts string class names to constants

@jbelien if there are too many changes for one PR I could split it


jbelien (approved):

jbelien:
Thanks a lot @fbuchlak 👍

PR #1253, opened 2025-03-23
feat: [Photon] support multiple layers

### LINKED ISSUE #1249: [Photon] missing support for multiple layers
#1245 introduced `layer` parameter for photon.

Photon supports filtering by [multiple layers](https://github.com/komoot/photon?tab=readme-ov-file#filter-results-by-layer).

https://photon.komoot.io/reverse?lat=49.001831&lon=21.239311&layer=city&limit=5
https://photon.komoot.io/reverse?lat=49.001831&lon=21.239311&layer=city&layer=district&limit=5

However it's not possible to filter with multiple layers with `Geocoder\Provider\Photon\Photon`.

fbuchlak:
Resolves #1249


src/Provider/Photon/Photon.php
```diff
@@ -177,6 +177,22 @@ public function getName(): string
...
+
+        if (!is_array($layers)) {
+            $layers = iterator_to_array($layers);
+        }
+
+        return implode('', array_map(
+            static fn ($layer) => sprintf('&layer=%s', $layer),
+            array_filter($layers, static fn ($layer) => is_scalar($layer)),
+        ));
+    }
```
jbelien, line 194: I find the way this function is written a bit confusing.
We indeed need to be able to process both `string` and `array` to avoid having a breaking change.

Couldn't we replicate what we did in `buildOsmTagFilterQuery()` function ?

```suggestion
    /**
     * @param string|string[] $layers
     */
    private function buildLayerFilterQuery(mixed $layers): string
    {
        if (is_string($layers)) {
            return '&layer='.urlencode($layers);
        }
        
        $query = '';
        foreach ($layersas $layer) {
            $query .= '&layer='.urlencode($layer);
        }
        return $query;
    }
```

jbelien (changes requested):

jbelien (approved):

